PR #436, opened 2015-01-16
name_or_id clean up for MultiTag

asobolev:
TODO: determine whether removeReference should also accept name

### DIFF
--- a/test/TestMultiTag.cpp
+++ b/test/TestMultiTag.cpp
@@ -171,6 +171,11 @@ void TestMultiTag::testReferences(){
     dt.addReference(da_2);
     CPPUNIT_ASSERT_MESSAGE(counterrmsg.str(), dt.referenceCount() == 2);
 
+    std::stringstream haserrmsg;
+    haserrmsg << "TestMultiTag::testReference: Has method did not work!";
+    CPPUNIT_ASSERT_MESSAGE(haserrmsg.str(), dt.hasReference(da_1.id()));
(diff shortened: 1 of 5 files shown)


test/TestMultiTag.cpp
```diff
@@ -171,6 +171,11 @@ void TestMultiTag::testReferences(){
     dt.addReference(da_2);
     CPPUNIT_ASSERT_MESSAGE(counterrmsg.str(), dt.referenceCount() == 2);
 
+    std::stringstream haserrmsg;
+    haserrmsg << "TestMultiTag::testReference: Has method did not work!";
```
gicmo: Couldn't that just be `std::string haserrmsg = "TestMultiTag::testReference: Has method did not work!"; `

asobolev: Coulb well be. But there is this stream everywhere, so I don't want to mess it up

jgrewe: That was the old style before we switched to use the CPPUNIT_ASSERT without message. 
Actually we could simply ```` CPPUNIT_ASSERT_MESSAGE("some message", x == y)```` if we want to keep the messages. 
imho, we can live with that style.  I kind of like the messages but probably they are more work than they are worth.

asobolev: I'd maybe keep it consistent - having stream for now and then change it to CPPUNIT_ASSERT_MESSAGE("some message", x == y) everywhere, if it is worth so. But then maybe in the separate PR.

PR #439, opened 2015-01-20
Group: convert to plain C HDF5 API

gicmo:


stoewer:
Nice! Maybe it's a good Idea to get rid of H5::File as well. If everyone is OK with that, I will do this on Friday.

stoewer:
Should I merge it or is this pr supposed to fix the errors on AppVeyor?

gicmo:
No, AppVeyor seems to be a different, i.e. H5 lib, issue; so go ahead an merge it ;-)

PR #451, opened 2015-01-29
Appveyor: ctest script

gicmo:


gicmo:
Windows is now included in the CDash page: http://0xdeadbabe.info/index.php?project=nix&date=2015-01-29

Issue #354, opened 2014-08-21
adding front end methods to access entities by name


jgrewe:
So far all methods like has-, get-, removeXYZ work with the id, now that we switched to the name being the identifying attribute, we should at least offer the methods to access entities by name.

jgrewe:
Apparently this issue can be closed.

PR #464, opened 2015-02-05
name_or_id clean-up for Section/Source

asobolev:


jgrewe:
does this imply that we can close issue #354? That would be great.

asobolev:
from what I have found - yes)

PR #470, opened 2015-02-07
fix wrong splitting of compound units

jgrewe:


src/util/util.cpp
```diff
@@ -151,18 +151,47 @@ void splitUnit(const string &combinedUnit, string &prefix, string &unit, string
 }
 
 
+void invertPower(std::string &unit) {
+    string p, u, pow;
```
gicmo: `pow` is a function, maybe better to call it something else.

PR #477, opened 2015-02-11
[test] Validate: also call file.validate()

gicmo:


gicmo:
`TestValidate::test` actually fails on windows. needs investigation.

PR #481, opened 2015-02-15
Extended some tests + warning fixes

gicmo:


gicmo:
+1 for Jan's changes from my part. ;-)

PR #489, opened 2015-02-27
Method to infer DataType from string (needed for nix-mx)

asobolev:


src/DataType.cpp
```diff
@@ -42,6 +44,35 @@ std::string data_type_to_string(DataType dtype) {
...
+            (*this)["int32"] = DataType::Int32;
+            (*this)["int64"] = DataType::Int64;
+            (*this)["uint8"] = DataType::UInt8;
+            (*this)["uint16"] = DataType::UInt16;
+            (*this)["uint32"] = DataType::UInt32;
+            (*this)["uint64"] = DataType::UInt64;
+            (*this)["string"] = DataType::String;
+            (*this)["nothing"] = DataType::Nothing;
+        }
+    };
```
jgrewe: in the reverse function (data_type_to_string) the names are all upper case. I guess we should have that consistent or handle it case-insensitive

asobolev: Thanks for the comment. So here the idea was to make it user/system convenient and accept all cases like "Double", "double", "DOUBLE" whatever "douBLe" etc. Such that, for instance, the Matlab "double" doesn't need to be converted to "Double" as it's clear it's a double type. Otherwise one has to go always do to the NIX code and see how exactly the string representation should be specified. What do you think?

src/DataType.cpp
```diff
@@ -42,6 +44,35 @@ std::string data_type_to_string(DataType dtype) {
     return str;
 }
 
+DataType string_to_data_type(std::string dtype) {
+
+    struct data_type_map : public std::map<std::string, DataType> {
```
stoewer: Don't inherit from STL containers - especially when this is not necessary.

I would rather use something like this (if you really want to use a map):
```c++ 
static std::map<std::string, DataType> type_map = {
    {"bool", DataType::Bool},
    {"char", DataType::Char},
    ...
}
```

src/DataType.cpp
```diff
@@ -42,6 +44,35 @@ std::string data_type_to_string(DataType dtype) {
     return str;
 }
 
+DataType string_to_data_type(std::string dtype) {
```
stoewer: Consider passing dtype as const reference.

stoewer:
About exceptions: `data_type_to_size()` throws `std::invalid_argument`. Should we do the same for `string_to_data_type()` and `data_type_to_string()`. The latter returns a string "FIXME" as default case: maybe this is not the best solution :-).

src/DataType.cpp
```diff
@@ -42,6 +44,31 @@ std::string data_type_to_string(DataType dtype) {
...
+        {"uint32", DataType::UInt32},
+        {"uint64", DataType::UInt64},
+        {"string", DataType::String},
+        {"nothing", DataType::Nothing}
+    };
+
+    std::string dtype_l = dtype;
+    std::transform(dtype_l.begin(), dtype_l.end(), dtype_l.begin(), ::tolower);
+
+    return type_map[dtype];
```
stoewer: 1. I guess this should be `return type_map[dtype_1]` and not `return type_map[dtype]`
2. If `dtype_1` is not a key in the map this actually inserts a new element [1], which is not intended.

[1] http://www.cplusplus.com/reference/map/map/operator[]

PR #492, opened 2015-03-11
TestUtil: fix warnings for msvc

stoewer:


test/TestUtil.cpp
```diff
@@ -138,7 +138,7 @@ void TestUtil::testConvertToSeconds() {
     int64_t m_value = 25;
     CPPUNIT_ASSERT_DOUBLES_EQUAL(1530.0, util::convertToSeconds(unit_min, min_value), std::numeric_limits<double>::round_error());
     CPPUNIT_ASSERT_DOUBLES_EQUAL(44100.0, util::convertToSeconds(unit_h, h_value), std::numeric_limits<double>::round_error());
-    CPPUNIT_ASSERT_DOUBLES_EQUAL(1500, util::convertToSeconds(unit_min, m_value), std::numeric_limits<double>::round_error());
+    CPPUNIT_ASSERT(1500 == util::convertToSeconds(unit_min, m_value));
```
jgrewe: I guess this should be 1500. or 1500.0 to avoid the warning...

stoewer: The warning was actually caused by the return value of `convertToSeconds()`, because it corresponds to the type of its second parameter - which in this case is `int64_t`. On the other hand this means that only integer types are compared and no `CPPUNIT_ASSERT_DOUBLES_EQUAL` is needed here.

PR #499, opened 2015-06-06
add NDSizeBase constructor that accepts vector arguments

sujithvm:
Added a constructor to NDSizeBase that accepts vector arguments. Thank you

### DIFF
--- a/include/nix/NDSize.hpp
+++ b/include/nix/NDSize.hpp
@@ -104,6 +104,25 @@ public:
         #endif
     }
 
+    template<typename U>
+    NDSizeBase(const std::vector<U> &args)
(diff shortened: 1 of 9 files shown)


sujithvm:
@stoewer Please review this pull request. Thank you

include/nix/NDSize.hpp
```diff
@@ -104,6 +104,25 @@ class NDSizeBase {
         #endif
     }
 
+    template<typename U>
+    NDSizeBase(std::vector<U> args)
```
gicmo: We shouldn't make a copy of the vector argument, i.e. pass the args as const-ref (`const std::vector<U> &args`)

PR #503, opened 2015-06-16
[helper.hpp] add empty constructor for struct Message

sujithvm:
Added an empty constructor for struct Message. Thank you


sujithvm:
JavaCPP throws an error '/usr/include/c++/4.8/bits/stl_construct.h:75:7: error: no matching function for call to ‘nix::valid::Message::Message()’ ' while generating a binding for std::vector<nix::valid::Message> . Hence this fix. Thank you

Issue #498, opened 2015-05-30
implicit use of a SetDimension as "ValueDimension"


jgrewe:
related to #398. So far we do not support the easy retrieval of event data which is used as a reference in a Tag.

Proposed solution:
The event data is comonly stored as a vector e.g. of event times with the only dimension being a Set dimension without any labels stored in it. Certain segments of the event trace are annotated with a Tag.
During retrieval of the tagged events (those that fall within the bounds established by position and extent) we return a DataView on the respective segment **assuming that they are ordered**.
Further we assume that position and extent have to interpreted as indices if the Set contains labels. If there are no labels, we apply them to the values.

Problem:
The values in the event trace, along a Set-Dimension, are not necessarily sorted.  Though I think that the unordered case is possible but rather unlikely, we may consider to improve the DataView to handle point selections.

jgrewe:
solved with pr #508

Issue #398, opened 2014-10-07
relation between entities we represent as multiTags


jgrewe:
The following problem: 
Consider the case in which we record a bunch of traces.
Then we have parts in these traces during which a certain condition was on, e.g. a stimulus was presented. This is represented as a MultiTag (Tag A) containing starts and extents of the stimulus presentations and that references the recorded traces.
Further assume that we detect certain events, let's say peaks, in the traces. There will be a MultiTag (Tag B) linking the times of the peaks to the raw trace in which they were detected. 
Now I want to have the times of the peaks that were detected while the stimulus was on.
More abstract, I would like to link the two tags. One could think that referencing the peak times in A would solve the problem but it does not. The positions and extents in A do not apply to the peak times ...

There are several approaches: 

1. Tag A and B both reference the trace in which the peaks were detected so there is an indirect link already. We could support these indirect links via filters on the API side. 
This solves the finding but I still cannot apply the position and extent of A to the peak times.

2. We could introduce a further dimension which tells that it is the data stored in the data array to which the position and extent have to be applied not a dimension descriptor.

jgrewe:
close via #508

PR #508, opened 2015-07-20
Use of DataArray data as ticks in RangeDimension

jgrewe:
This pr should solve #498, (and parts of #398).
The data stored in a 1-d dataArray is used as the ticks of a RangeDimension. It is called AliasRangeDimension (we can discuss better names ;-), I would appreciate better ideas).
There are certain restrictions on this:
1. the data must be 1-D
2. the data must be numeric

Since this involves quite some changes I would really appreciate a critical review.

### DIFF
--- a/include/nix/DataArray.hpp
+++ b/include/nix/DataArray.hpp
@@ -377,14 +377,17 @@ public:
      */
     RangeDimension createAliasRangeDimension() {
         if (this->dataExtent().size() > 1) {
-            throw nix::InvalidDimension("AliasRangeDimensions oly allowed for 1D numeric DataArrays!",
-                                        "DataArray::appendAliasRangeDimension");
+            throw nix::InvalidDimension("AliasRangeDimensions only allowed for 1D numeric DataArrays!",
+                                        "DataArray::createAliasRangeDimension");
+        }
+        if (!nix::data_type_is_numeric(this->dataType())) {
+            throw nix::InvalidDimension("AliasRangeDimensions are only allowed for 1D numeric DataArrays!",
+                                        "DataArray::createAliasRangeDimension");
         }
         if (dimensionCount() > 0) {
             throw nix::InvalidDimension("Cannot append additional alias dimension. There must only be one!",
-                                        "DataArray::appendAliasRangeDimension");
+                                        "DataArray::createAliasRangeDimension");
         }
-        // TODO:: test for numeric data types
         return backend()->createAliasRangeDimension();
     }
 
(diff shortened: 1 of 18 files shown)


src/DataType.cpp
```diff
@@ -74,6 +74,15 @@ DataType string_to_data_type(const std::string& dtype) {
     return type_map[dtype_l];
 }
 
+
+bool data_type_is_numeric(DataType dtype) {
+    return !(dtype == DataType::Bool ||
+             dtype == DataType::Char ||
+             dtype == DataType::String ||
+             dtype == DataType::Nothing);
+}
```
gicmo: `::Opaque`, `::DateTime` and `Date` should probably be in that list too, no? Probably saver to whitelist because if we extend the list and forget to add it here we might be in trouble.

test/TestDataSet.cpp
```diff
@@ -227,6 +226,34 @@ void TestDataSet::testDataTypeFromString() {
...
+            {true, nix::DataType::UInt16},
+            {true, nix::DataType::Int32},
+            {true, nix::DataType::UInt32},
+            {true, nix::DataType::Int64},
+            {true, nix::DataType::UInt64},
+            {true, nix::DataType::Float},
+            {true, nix::DataType::Double},
+            {false, nix::DataType::String},
+            {false, nix::DataType::Nothing}
+    };
```
gicmo: Not all DataTypes covered here.

gicmo:
Awesome work!

include/nix/DataArray.hpp
```diff
@@ -313,6 +313,18 @@ class NIXAPI DataArray : public base::EntityWithSources<base::IDataArray>, publi
         }
         return backend()->createRangeDimension(backend()->dimensionCount() + 1, ticks);
     }
+    
+    /**
+     * @brief Append a new RangeDimension that uses the data stored in this DataArray as ticks.
+     * This works only(!) if the DataArray in 1D and the stored data is numeric. An Exception
```
gicmo: Maybe specify which exception will be thrown (i.e. `InvalidDimension` exception) so one can catch the correct one.

test/TestDataSet.cpp
jgrewe: uh, right. Will be fixed.
They should also appear in the DataType string_to_data_type and its companion, shouldn't they?

gicmo: Oh, yeah, I guess so. At least `Opaque`. `Date+DateTime` are never used anywhere. Maybe we should kill them?

jgrewe:
just re-based the branch, should be smoothly mergeable ;-)

Issue #495, opened 2015-03-18
Public value attributes are not persisted


stoewer:
In the current version all public attributes of `nix::Value` such as `uncertainty`, `checksum` etc are not persisted in the file. The following code ...

```c++
File f = File::open("test_file_test9.h5", FileMode::Overwrite);

Section s = f.createSection("foo", "footype");
Value v{10};
v.uncertainty = 20;
v.checksum = "checksum";
v.reference = "reference";
v.encoder = "encoder";
v.filename = "filename";
Property p = s.createProperty("val", v);

Value v2 = p.values()[0];

cout << "v2.get(): " << v2.get<int>() << endl;
cout << "v2.uncertainty: " << v2.uncertainty << endl;
cout << "v2.checksum: " << v2.checksum << endl;
cout << "v2.reference: " << v2.reference << endl;
cout << "v2.encoder: " << v2.encoder << endl;
cout << "v2.filename: " << v2.filename << endl;

f.close();
```
... produces this output:

```
v2.get(): 10
v2.uncertainty: 0
v2.checksum: 
v2.reference: 
v2.encoder: 
v2.filename: 
```

When opening the file with h5view one can see that the respective dataset does not contain the missing attribute values. 

There is also no test covering the `nix::Value` attributes.

gicmo:
Should be fixed with pr #509.

PR #509, opened 2015-07-22
Value bug

jgrewe:
collaborative fix of bug #495


stoewer:
I restarted the CI twice but still coveralls is not working. Should we merge it anyway?

jgrewe:
wait for a moment, I will add the respective test cases and then we will see again

Issue #473, opened 2015-02-10
[m32] ndsize_t != size_t related issues on 32bit


gicmo:
On platforms where `size_t` is smaller then `ndsize_t` (i.e. win 32bit) we have a few places where we return `size_t` instead of `ndsize_t` or where we don't handle overflows...

- [x] `NDSize::nelms()`
    - [x] `NDArray::allocate`
- [x] `hdf5::Group::objectCount`
  - [x] `BaseTag::referenceCount`
  - [x] `BaseTag::featureCount`
  - [x] `Block::dataArrayCount`
  - [x] `Block::multiTagCount`
  - [x] `Block::sourceCount`
  - [x] `Block::tagCount`
  - [x] `DataArray::dimensionCount`
  - [x] `EntityWithSources::sourceCount`
  - [x] `File::blockCount`
  - [x] `File::sectionCount`
  - [x] `Section::propertyCount`
  - [x] `Section::sectionCount`
  - [x] `Source::sourceCount`
  - [x] `Property::valueCount`
- [x] `hdf5::Group::objectName`
  - [x] `BaseTag::getReference`
  - [x] `BaseTag::getFeature`
  - [x] `Block::getDataArray`
  - [x] `Block::getMultiTag`
  - [x] `Block::getSource`
  - [x] `Block::getTag`
  - [x] `EntityWithSources::sourceCount`
  - [x] `File::getBlock`
  - [x] `File::getSection`
  - [x] `Section::getProperty`
  - [x] `Section::getSection`
  - [x] `Source::getSource`

gicmo:
This should be solved now that pr #545 is merged.

PR #512, opened 2015-07-23
Several changes from size_t to using of ndsize_t...

jgrewe:
done for Block, Source, Section, and File (cf. #473)


gicmo:
Awesome!

PR #513, opened 2015-07-24
[<< operators] in Section and Source reverted back to pre pr #512 state

jgrewe:
with pr #512 I changed the << operators to the same version as in all other entities. With these changes, nixpy would compile but tests failed during nix import due to undefined symbols.
This commit reverts these changes. Still, I do not really understand but this may come later ...


gicmo:
Oh man sorry, should have spotted that.

PR #529, opened 2015-09-09
Frontend and HDF5 back-end changes from the file_backend

jgrewe:


src/hdf5/FeatureHDF5.cpp
```diff
@@ -85,20 +85,13 @@ void FeatureHDF5::data(const std::string &name_or_id) {
 
 shared_ptr<IDataArray> FeatureHDF5::data() const {
     shared_ptr<IDataArray> da;
-    bool error = false;
 
     if (group().hasGroup("data")) {
         Group other_group = group().openGroup("data", false);
         da = make_shared<DataArrayHDF5>(file(), block, other_group);
-        if (!block->hasDataArray(da->id())) error = true;
+        if (!block->hasDataArray(da->id()))
```
gicmo: Shouldn't we stay consistent and have:

```c++
if (!block->hasDataArray(da->id())) {
    throw std::runtime_error("FeatureHDF5::data: DataArray not found!");
}
```

src/Section.cpp
```diff
@@ -47,7 +47,7 @@ Section::Section(shared_ptr<base::ISection> &&ptr)
 
 
 void Section::link(const Section &link) {
-    if (link == none) {
+    if (!link) {
         backend()->link(none);
     } else {
         backend()->link(link.id());
```
gicmo: I would make the whole thing into `backend()->link(link ? link.id() : none);`

src/hdf5/FeatureHDF5.cpp
jgrewe: do you mean the curly braces?

gicmo: Yeah. ;)

src/Section.cpp
jgrewe: no match for ternary ‘operator?:’ (operand types are ‘bool’, ‘std::string {aka std::basic_string<char>}’, and ‘const none_t {aka int boost::detail::none_helper::* const}’)

gicmo: Oh! na, has to be the same type. Sure. Sorry.

gicmo:
LGTM

Issue #523, opened 2015-09-09
Back-end: distinguish between creation and opening of entities


stoewer:
Currently the back end classes (such as BlockHDF5) have constructors used for both, creation and opening of entities.
It would be nicer to used constructors just for opening entities and create them using a static member function `create`.

This is also related to #305

stoewer:
I just looked at the current implementation of our `createSomething()` methods in the back-end. A typical example looks like this:

```c++
shared_ptr<base::IBlock> FileHDF5::createBlock(const string &name, const string &type) {
    if (name.empty()) {
        throw EmptyString("Trying to create Block with empty name!");
    }
    if (hasBlock(name)) {
        throw DuplicateName("Block with the given name already exists!");
    }
    string id = util::createId();

    Group group = data.openGroup(name, true);
    return make_shared<BlockHDF5>(file(), group, id, type, name);
}
```

There are two things I'd like to highlight here:

1. The check `if(name.empty())` is also implemented in the ctor of `NamedEntityHDF5` which is used for creation of entities -> the ctor throws the same exception.
2. We do everything to prevent the exceptions thrown by ctors because we do not handle them in our `createSomething()` methods (see last line in the listing above).

I therefore suggest the following changes:

1. Implement a static member function `init` or `init_group` for all back-end entities. Those functions initialize a newly created group before it is passed to a ctor. They should do all checks and throw exceptions if necessary.
2. Make ctors exception free.
3. Implement all `createSomething()` methods like this:

```c++
shared_ptr<base::IBlock> FileHDF5::createBlock(const string &name, const string &type) {
    if (hasBlock(name)) {
        throw DuplicateName("Block with the given name already exists!");
    }

    Group group = data.openGroup(name, true);

    try {
        BlockHDF5::init(group, util::createId(), type, name)
        return make_shared<BlockHDF5>(file(), group);
    } catch(exception e) {
        data.removeGroup(name);
        throw e;
    }
}
```

jgrewe:
Would you then do all the subgroup creation (e.g. for features in Tag or dimensions in DataArray etc) do in the init method?
Why delegation to the init method? Wouldn't it be enough to do that in the create method?

stoewer:
> Would you then do all the subgroup creation (e.g. for features in Tag or dimensions in DataArray etc) do in the init method?

Yes, especially everything that can go wrong.

> Why delegation to the init method? Wouldn't it be enough to do that in the create method?

[...]

Issue #305, opened 2014-07-27
DataTag ctor throwing errors


balint42:
Currently the `DataTag` ctor will throw errors if an invalid `DataTag` is opened: it uses the `positions` method to set the supplied obligatory `DataArray` as `positions` array. This is not a good situation, as the error will obstruct every other operation (e.g. validation) one could want to perform on the `DataTag`. Thus we should perhaps consider not doing any kind of check in the ctor and set the supplied `DataArray`'s `id` no matter what.

PR #530, opened 2015-09-10
Ctor vs init

stoewer:
This pr serves the purpose of discussing the changes related to #523 and is not necessarily meant to be merged.

As an example I implemented `init` as static member function for some base classes and used them to implement `FileHDF5::createBlock` as discussed in #523. But there are certain things that need to be discussed:

* So far we have a rather stupid front-end and the majority of front-end methods just call `backend()->something()` and that's it. But with #529 and this pr we are changing the paradigm a bit since we do more checks in the front-end and rely in the back-end on more or less sane parameters. Is this the direction we want to go?
* Using a static member functions for `init` has one disadvantage. Since `init` methods have different signatures it is possible to call different `init` methods from `BlockHDF5`: the one defined by `EntityHDF5` and the one on `NamedEntityHDF5`, but only the latter is the correct one. This is not very nice.

### DIFF
--- a/include/nix/File.hpp
+++ b/include/nix/File.hpp
@@ -154,6 +154,15 @@ public:
      * @return The created block.
      */
     Block createBlock(const std::string &name, const std::string &type) {
+        if (name.empty()) {
+            throw EmptyString("Trying to create Block with empty name!");
+        }
+        if (type.empty()) {
+            throw EmptyString("Trying to create Block with empty type!");
+        }
+        if (hasBlock(name)) {
+            throw DuplicateName("Block with the given name already exists!");
+        }
         return backend()->createBlock(name, type);
     }
 
--- a/include/nix/hdf5/BlockHDF5.hpp
+++ b/include/nix/hdf5/BlockHDF5.hpp
@@ -39,29 +39,6 @@ public:
      */
     BlockHDF5(const std::shared_ptr<base::IFile> &file, const Group &group);
 
-    /**
-     * Standard constructor for a new Block.
-     *
-     * @param file      The file which contains this block.
-     * @param group     The group that represents the block inside the file.
-     * @param id        The id of this block.
-     * @param type      The type of this block.
-     * @param name      The name of this block.
-     */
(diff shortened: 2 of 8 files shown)


jgrewe:
hm, pr did not change much (except, perhaps, for DataArray.cpp where the existing check was changed a bit) in the front-end classes. I would think that moving logic, that works with front-end methods only (e.g. does not need access to the group/directories directly) is desirable when we avoid code-duplication. Do you see a situation in which one might use the back-ends independent of the front-ends? This would be an argument against error checking in the front-end.

Regarding the init method I would argue in a similar way. If we do not have a use-case in which we have users accessing creating their own objects directly, I see no flaw in the design. It is not fail safe, I agree, but only during development time of a back-end.

gicmo:
Re frontend-backend: I though we always wanted to have more logic in the frontends but, I guess mostly because there was no other backend, we never had the need to actually do it. Anyway, while certainly a change to the current status-quo, I welcome it.

Re init: Having actually seen the implementation, I am liking the idea less, although I still dislike the two highly confusable constructor that look very similar, but do totally different things more (little side note: a prototypical example of 11 lines of useless "documentation" for the second `Block` constructor. ;-)). Although not changed with the new implementation I also don't like it from an aesthetics kind of view that the group+id are created by the parent entity and the init is then done by the new child entity. I feel that should be long together.
The try-catch change is anyway independent of these changes here and necessary because of issue #532.

stoewer:
> Re init: Having actually seen the implementation, I am liking the idea less, although I still dislike the two highly confusable constructor that look very similar

OK, I'll try to come up with some alternative approach for the static `init` member.

stoewer:
I tried to come up with an alternative to the static `init` members and played around with some other solutions, but they all have their own flaws. If someone has a good idea how to solve #305 and/or #523 I am more than happy to implement it. If not I suggest to close this pr and move #305 and #523 to M 1.x.

gicmo:
I would say lets move #305 and #523 to M 1.x, because I would really like to have a 1.1 release soon, so I can try to get our formula into home-brew science :)

PR #533, opened 2015-09-16
DataType: remove Date, DateTime

gicmo:
Not supported anyway. Fix all locations that mention it.
NB: This will break nixpy and nix-java bindings.

### DIFF
--- a/include/nix/DataType.hpp
+++ b/include/nix/DataType.hpp
@@ -38,8 +38,6 @@ NIXAPI enum class DataType {
     UInt32,
     UInt64,
     String,
-    Date,
-    DateTime,
     Opaque,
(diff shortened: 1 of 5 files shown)


stoewer:
> This will break nixpy and nix-java bindings.

Is nix-mx also affected? We should file issues for all affected sub projects and add them to a 1.1 milestone.

gicmo:
@stoewer nope, nix-mx seems to be clean.

gicmo:
@stoewer could file issue, but as soon as this is merge I will fix the corresponding bindings, not sure if it necessary.

stoewer:
@gicmo Even better!

Issue #524, opened 2015-09-09
move checks to front-end


jgrewe:
checks of user-input that throw exceptions should go to the front-end to ensure similarity of checks

jgrewe:
this might/should be solved with mit #536

PR #534, opened 2015-09-17
Move checks to front-end ... 

jgrewe:
... first chunk of changes, please have a look.

edit:
Added the requested checkNameAndType method.

### DIFF
--- a/include/nix/Block.hpp
+++ b/include/nix/Block.hpp
@@ -15,7 +15,6 @@
 #include <nix/DataArray.hpp>
 #include <nix/MultiTag.hpp>
 #include <nix/Tag.hpp>
-
 #include <nix/Platform.hpp>
 
 #include <string>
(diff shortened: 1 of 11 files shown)


gicmo:
`checkName` and `checkType` are so often done in a pair, I think a `checkNameAndType` could nice.

gicmo:
Btw, probably not a big deal, but moving logic from the header to the source prevents inlining of that function.

include/nix/Block.hpp
```diff
@@ -15,7 +15,7 @@
 #include <nix/DataArray.hpp>
 #include <nix/MultiTag.hpp>
 #include <nix/Tag.hpp>
-
+#include <nix/util/util.hpp>
```
gicmo: Is that really needed? Especially since you moved the checks to the source file?

include/nix/File.hpp
```diff
@@ -13,7 +13,7 @@
 #include <nix/base/IFile.hpp>
 #include <nix/Block.hpp>
 #include <nix/Section.hpp>
-
+#include <nix/util/util.hpp>
```
gicmo: Same as above.

gicmo:
Part of issue #524

PR #536, opened 2015-09-20
Front end checks

jgrewe:
Move remaining checks to front-end. This should complete issue #524.


jgrewe:
so, i tried the "update branch" button, probably not so nice a feature since we now have the merge commit... If you prefer I am happy to rebase and submit it again.

Issue #542, opened 2015-10-17
License must be changed to e.g. BSD-3


yarikoptic:
BSD-4 used by the project ATM is not even compatible with GPL (and probably other copyleft licenses) due to inclusion of the advertisement clause (#3).  Such BSD licenses were dropped even by original users of them, see http://www.gnu.org/licenses/license-list.en.html#OriginalBSD

gicmo:
I guess BSD-4 this was a poor choice then because we clearly don't want to be incompatible with GPL. I guess the closest alternative would be [Modified BSD](http://www.gnu.org/licenses/license-list.en.html#ModifiedBSD), unless someone else has a better idea.

gicmo:
We should fix this for 1.1.

jgrewe:
No objection, neither against, the version 1.1, nor the modified BSD

stoewer:
Also no objections from my side. I was not aware of the incompatibility issues between BSD 4-clause and GPL. Here is a good summary of GPL compatible and incompatible licenses:

https://www.gnu.org/licenses/license-list.en.html

I also think the modified BSD (3-clause) license would be a good choice. The above site further suggests Apache 2 as an alternative.

stoewer:
Maybe we should use this opportunity to update the copyright holder list in the head of our license file.

balint42:
I think that BSD (3-clause) seems totally valid! Yes to BSD 3. :-)

stoewer:
In our source files we have quite a mix of copyright and author declarations. In most of the files, even those with contributions from many authors, we have just the following copyright notice:

```c
// Copyright (c) 2013, German Neuroinformatics Node (G-Node)
...
```

In some others we have multiple copyright holders such as:

```c
// Copyright © 2015 German Neuroinformatics Node (G-Node)
//                  John Doe <john.doe@example.org>
...
```

To make it even more diverse we also have headers like this:

```c
// Copyright © 2015 German Neuroinformatics Node (G-Node)
...
// Author: John Doe <john.doe@example.org>
```

If we change our license it might be a good idea to unify our license headers too. In other projects I have seen something like this:

```c
// Copyright © 2015 German Neuroinformatics Node (G-Node),
//                  Contributors as listed in Contributors.md
...
```

The `Contributors.md` file contains a list of all contributors with name and mail address. IMHO this is a very practical way to acknowledge all contributions (and the inherent moral rights of authors).

* Is such a copyright delegation to a contributors file legally OK?
[...]

Issue #403, opened 2014-10-18
upload to Debian proper and NeuroDebian?


yarikoptic:
Hi guys,

I wondered if you would be interested to get nix uploaded to Debian proper and NeuroDebian (complimenting your PPA with builds for Debians as well)?

few points upon quick look:
- debian/control
  - adjust Vcs *and homepage) field to point to nix instead of pandora
  - add Balint Morvai balint@morvai.de to Uploaders to pacify lintian
- debian/copyright
  - convert to http://www.debian.org/doc/packaging-manuals/copyright-format/1.0/  format
  - list h5py license as pertinent to ./src/hdf5/DataSet.cpp 
- ITP issue would need to be filed and closed in the latest changelog entry
- previous changelog entries ideally should be marked as 'UNRELEASED' since they were not uploaded to any of those now listed releases of ubuntu

I would also recommend to 
- generate debian/lib*.symbols file -- see e.g. https://wiki.debian.org/UsingSymbolsFiles

I will run the build now to see if anything new pops out

cheers

PR #543, opened 2015-10-17
ENH: debian packaging DEP5 copyright file, boost policy, etc

yarikoptic:
Apparently I have already suggested/whined a while back (#403) so while at the meeting decided to look/help, thus this PR.  Note/address #542 as well please.


yarikoptic:
oh boy - that is an obscure coveralls failure -- do you guys aim for decreased coverage? ;)

gicmo:
I guess the issue is that although it increased on this branch, the coverage is still lower compared to master. I think we can ignore that here since we just need to merge master.

gicmo:
Merged via command line, also issue #542 is fixed now so I have updated debian/copyright to reflect this. Thanks again!

PR #544, opened 2015-10-18
[m32] Use ndsize_t for object indexing everywhere now

gicmo:
The last bits of issue #473. Greetings from SfN. ;)


gicmo:
This needs some more fixes for the dimensionCount related changes. I will be fixing it ...

Issue #526, opened 2015-09-09
Explore solutions for multi back-end unit tests


stoewer:
Explore solutions for multi back-end unit tests

stoewer:
Just found a stackoverflow question about parameterized tests with cpp-unit:

https://stackoverflow.com/questions/290099/parameterizing-a-test-using-cppunit

jgrewe:
looks interesting but introduces a further, I will try it the next days

stoewer:
I just tested two different ways how unit tests can be executed for different back-ends. Sadly both variants only work if the back-ends to test are already known at compile time. I implemented them in a small toy project.

Here is the class under test I used in my small toy example:
```c++
class Foo {
    std::string foo;
public:

    Foo(const std::string foo);
    std::string get_foo() const;
};
```

The first version I tried uses inheritance in order to create test suites that have their own setUp() and tearDown() methods. Here is how this looks like:

```c++
class TestFooInherit : public CPPUNIT_NS::TestFixture {

protected:
    Foo foo;

public:
    TestFooInherit()
            : foo("")
    {}

    void test_foo() {
        CPPUNIT_ASSERT(foo.get_foo().size() > 0);
        std::cout << std::endl << foo.get_foo() << std::endl;
    }
};


class TestFooInherit1 : public TestFooInherit {

    CPPUNIT_TEST_SUITE(TestFooInherit1);
        CPPUNIT_TEST(test_foo);
    CPPUNIT_TEST_SUITE_END();

public:
    void setUp() {
        foo = Foo("blafasel");
    }

    void tearDown() {}
};

class TestFooInherit2 : public TestFooInherit {

    CPPUNIT_TEST_SUITE(TestFooInherit2);
        CPPUNIT_TEST(test_foo);
    CPPUNIT_TEST_SUITE_END();

public:
    void setUp() {
        foo = Foo("somethingelse");
    }

    void tearDown() {}
};
```

Since the actual test suite is defined in the child classes this method allows to disable tests of unimplemented featured during the development of a back-end.

Here is the other version that works with templates:
```c++
struct Param1 {
    std::string param = "balfasel";  // this would be our back-end name in NIX
};

struct Param2 {
    std::string param = "somethingelse";
};


template<class PARAM>
class TestFooTemplate : public CPPUNIT_NS::TestFixture, public PARAM {

    CPPUNIT_TEST_SUITE(TestFooTemplate<PARAM>);
        CPPUNIT_TEST(test_foo);
[...]

PR #546, opened 2015-11-02
Back-end specific tests via inheritance

stoewer:
This small pr show how to run the same tests multiple back-ends as discussed here #526 . As a start I just changed the test for `TestFile`. But if everyone is OK with this kind of pattern for our tests I'll apply it to other tests as well.


stoewer:
This is just the next bunch of tests. There will be more.

stoewer:
Now all relevant tests can be subclassed in order to implement tests for multiple back-ends

Issue #556, opened 2015-11-22
Value/Variant: make internal representation better compatible for H5 IO


gicmo:
For H5 IO: `std::string` needs to be `const char *` and `bool` should be `unsigned char`.

gicmo:
Bool should be represented as enum now in the file (#635), which makes it easy to just write bools as in bool. So this is fixed.

PR #557, opened 2015-11-22
[Value] internal representation for std::string is now char *

gicmo:
Issue #556


gicmo:
Bool should be represented as enum now in the file, which makes it easy to just write bools as in `bool`. So this is fixed.

Issue #554, opened 2015-11-21
Add the concept of a Group to the data model


stoewer:
A Group is an additional grouping element inside Blocks. Entities can be in multiple groups.

gicmo:
We should probably update the [wiki](https://github.com/G-Node/nix/wiki/The-Model) too.

jgrewe:
This issue can be closed, the wiki has been updated,

PR #558, opened 2015-11-22
implementation of a simple group

jgrewe:
the implementation is really dump and repetitive... I wonder if some template magic would be nice at this point? Comments are welcome!

solves issue  #554

### DIFF
--- a/src/hdf5/GroupHDF5.cpp
+++ b/src/hdf5/GroupHDF5.cpp
@@ -114,12 +114,12 @@ void GroupHDF5::dataArrays(const std::vector<DataArray> &data_arrays) {
 
     // extract vectors of names from vectors of new & old references
     std::vector<std::string> names_new(data_arrays.size());
-    transform(data_arrays.begin(), data_arrays.end(), names_new.begin(), util::toName<DataArray>);
-    //FIXME: issue 473
+    std::transform(data_arrays.begin(), data_arrays.end(), names_new.begin(), util::toName<DataArray>);
+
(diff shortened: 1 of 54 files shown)


src/hdf5/GroupHDF5.cpp
```diff
@@ -0,0 +1,353 @@
...
+#include <nix/Block.hpp>
+#include <nix/hdf5/SourceHDF5.hpp>
+#include <nix/hdf5/DataArrayHDF5.hpp>
+#include <nix/hdf5/TagHDF5.hpp>
+#include <nix/hdf5/MultiTagHDF5.hpp>
+
+#include <boost/range/irange.hpp>
+#include <nix/hdf5/BlockHDF5.hpp>
+
+using namespace std;
```
gicmo: Booooo! :-)

jgrewe: right you are...

src/hdf5/GroupHDF5.cpp
```diff
@@ -0,0 +1,352 @@
...
+    return removed;
+}
+
+
+void GroupHDF5::dataArrays(const std::vector<DataArray> &data_arrays) {
+
+    // extract vectors of names from vectors of new & old references
+    std::vector<std::string> names_new(data_arrays.size());
+    transform(data_arrays.begin(), data_arrays.end(), names_new.begin(), util::toName<DataArray>);
+    //FIXME: issue 473
```
gicmo: Actually issue #473 is fixed. oO

PR #561, opened 2015-11-24
DataType: check for _WIN32 instead of WIN32

stoewer:
* For consistency reasons
* And lets hope this causes less troubles with nix-java


stoewer:
I hate when I find the typos after creating the pr

Issue #565, opened 2015-12-05
strange behavior with deleted entities.


jgrewe:
I found the following code in the tests:
````
 l1n1.deleteSection(l2n2.id());
...
 CPPUNIT_ASSERT(!l1n1.hasSection(l2n2));
````
This looks cool, but is really strange. In the first line section l2n2 is deleted, and in the last line we ask whether it exists. ``Section::hasSection(section)`` does ``return backend()->hasSection(section.id());`` looks ok, but in this context it should fail! We deleted l2n2 from the file. The l2n2 variable, is not None, however. The strange thing is, that we can ask for its name, or id and get proper results, even though it was deleted just moments before. Why is this possible? Probably because hdf5 caches or does not flush between the two lines in the code. The previous test fails in the file-backend because l2n2 does no longer exist. 

Imho, the ``hasSection()``, and probably all other ``hasXYZ()`` functions should be changed to catch the possibility that the entity has been deleted but is not none. I do not like try catch statements that much, but it would be one option.

jgrewe:
solved via pull requests #566 and #567

PR #566, opened 2015-12-11
add a way to check whether an Entity is valid in the backend

jgrewe:
In hdf5 an hid identifier can be valid even if the object is unlinked from everything else in the file, the object's name, btw., does no longer exist at that point. According to the valid hid, there may still be a reference count from this object to attributes and datasets. These can still be accessed.

This pr is related to issue #565


gicmo:
LGTM!

PR #567, opened 2015-12-13
Improved Input checking

jgrewe:
Front facing classes check passed entities for none and if the are valid in the backend

This pr should solve issue #565


gicmo:
LGTM, @jgrewe feel free! ;)

Issue #594, opened 2016-02-03
Assignment operator of Dimension.cpp


jgrewe:
The assignment operator  in Dimension.cpp checks for the type of the passed Dimension and only assigns if types match. If they do not, it does not do anything. 

It should throw an exception...

PR #599, opened 2016-02-06
Exceptions for dimension assign operator #594

jgrewe:


src/Dimensions.cpp
```diff
@@ -221,8 +221,10 @@ SampledDimension& SampledDimension::operator=(const Dimension &other) {
 
     if (other.dimensionType() == DimensionType::Sample && impl() != tmp) {
         std::swap(impl(), tmp);
+    } else {
+        throw nix::IncompatibleDimensions("Cannot assign dimension of type " + nix::util::dimTypeToStr(other.dimensionType())
```
gicmo: Small thing: Now the exception in the else case will be thrown if it either not the right type, or if we have self-assignment.

PR #600, opened 2016-02-09
[hdf5] move Value writing code to Property

gicmo:
Just moving code to where it belongs.


backend/hdf5/PropertyHDF5.cpp
```diff
@@ -233,12 +444,6 @@ void PropertyHDF5::values(const nix::none_t t) {
 }
 
 
-bool PropertyHDF5::isValidEntity() const {
-    return dataset().referenceCount() > 0;
```
gicmo: Hmm, not sure why I removed it, but it seems nobody uses it anyway.

gicmo: Ok, didn't remove it as @jgrewe pointed out but just moved it. Doh!

PR #618, opened 2016-07-19
Use order tracking for H5Group::objectName

### LINKED ISSUE #617: H5Group::objectName does not use creation order index
Link creation order is tracked both at the file level and at the group level. This was introduced in PR #390 after a discussion in issue #387. However, when accessing groups by their index, using `H5Lget_name_by_idx`, the default index type and order is used, `H5_INDEX_NAME` and `H5_ITER_NATIVE` respectively (see [H5Group.cpp](https://github.com/G-Node/nix/blob/edad3cb1f90c0a7a2b6cbb9c612d984afe164ad8/backend/hdf5/h5x/H5Group.cpp#L135-L142)).

This is an incomplete solution and appears to retrieve groups in creation order only when the number of child groups is less than 8 (I am unsure why this is). For N > 8, the indexed retrieval order appears arbitrary (yet stable).

To make retrieval and iteration follow creation order, the index type and order in `H5Lget_name_by_idx` should be `H5_INDEX_CRT_ORDER` and `H5_ITER_INC` respectively.

achilleas-k:
Use correct index type and order in `H5Lget_name_by_idx`: `H5_INDEX_CRT_ORDER` and `H5_ITER_INC` respectively.

File and Group in TestH5Group updated to enable order tracking and a new test has been added that checks whether retrieval order follows creation order.

Resolves #617.

### DIFF
--- a/backend/hdf5/h5x/H5Group.cpp
+++ b/backend/hdf5/h5x/H5Group.cpp
@@ -134,8 +134,8 @@ std::string H5Group::objectName(ndsize_t index) const {
     // check whether name is found by index
     ssize_t name_len = H5Lget_name_by_idx(hid,
                                           ".",
-                                          H5_INDEX_NAME,
-                                          H5_ITER_NATIVE,
+                                          H5_INDEX_CRT_ORDER,
+                                          H5_ITER_INC,
                                           (hsize_t) index,
                                           NULL,
                                           0,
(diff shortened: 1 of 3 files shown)


jgrewe:
uh, I am afraid you have to rebase ...

achilleas-k:
Done.

Issue #593, opened 2016-02-03
Removing dimensions


jgrewe:
So far, we treat dimensions as a stack. Let's assume we have 2 dimensions in one DataArray. If the first one is removed the second one becomes the new number 1. That means, the group is renamed and the index is updated to 1. This is kind-of-strange behavior, especially since we do not support inserting of dimensions. 

As I see it, we have two options. 
1) We support inserting dimensions, this requires to replace the DataArray::appendDimension methods.
2) Allow setting and removing dimensions only in unison

There is one more issue: 
````
    double samplingInterval = boost::math::constants::pi<double>();
    Dimension d = data_array.appendSampledDimension(samplingInterval);
    Dimension d2 = data_array.appendSampledDimension(samplingInterval);
   
    data_array.deleteDimension(d.index());
    // at this point group and index of d2 in the backend should have changed.
    std::cerr << d2.index() << std::endl;
    // actually, the output will be 2 though the current index of d2 should be 1
    // the following command works with the hdf5 backend, but fails, correctly, in the fs backend.
    data_array.deleteDimension(d2.index());
````

sudoankit:
I want to work on this.
I am reading and checking the `fs backend`. Is it the right way?
Please guide how to start fixing this.

jgrewe:
This issue relates to both backends, not only the file-system. The tricky part is to find out which way to go.  I personally think, that we should not handle the dimensions as a stack. Rather the dimensions should keep their index and should not be renamed. 
This would need mechanism for the insertion and the deletion of selected dimensions.

jgrewe:
@sudoankit if you want to tackle this issue, this would be the way to go.

1. disallow the removal of individual dimensions, i.e remove the respective methods.
2. provide a method to remove all dimensions at once.
3. do not re-arrange remaining dimensions.
4. adapt test cases.
5. check both back-ends.

Please check out the Contributing.md file. If you need more info, let us know.

sudoankit:
Thank you so much. Danke schön.
I am understanding the classes and methods used in NIX and brushing up the boost library.
I am learning to write test cases. It's going well.
When in doubt can I ask you for help in the IRC chat or mail or here at Github? Do we have a gitter chat room for G-Node/NIX? I have read the contributing file.
Also, can I be assigned to this?

First, I will try to recreate and understand methods for dimensions.
I will update you with my work regularly. 

[out of context: the contributing file has a small typo in 'issue tracker section'. crated -> created]

[...]

PR #619, opened 2016-07-24
Dimensions are no longer treated as stack...

jgrewe:
handling issue #593:

removing of individual dimensions is no longer possible, respective method is gone. new deleteDimensions() removes all at once. 

public createDimension methods on DataArray are deprecated (documentation) and ignore any passed id argument.

### DIFF
--- a/test/TestValidate.cpp
+++ b/test/TestValidate.cpp
@@ -80,9 +80,7 @@ void TestValidate::setValid() {
         }
     }
     array4.setData(sin_array);
-    for (ndsize_t i = array4.dimensionCount(); i > 0 ; --i) {
-        array4.deleteDimension(i);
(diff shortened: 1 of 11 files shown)


test/TestValidate.cpp
```diff
@@ -348,6 +334,7 @@ void TestValidate::test() {
     CPPUNIT_ASSERT_EQUAL(false, myResult.hasErrors());
     
     myResult = file.validate();
+    std::cerr << myResult << std::endl;
```
gicmo: Debugging leftover or on purpose?

jgrewe: ups, indeed some debug code. Will remove it and the rebase, when the other pr is merged...

Issue #560, opened 2015-11-23
cleanup of repetitive code


jgrewe:
with pr #558 I introduced even some more repetitive code. We have several instances where we have sets of methods of the following pattern:
- [ ] `entityCount()`
- [ ] `getEntity()`
- [ ] `addEntity()`
- [ ] `hasEntity()`
- [ ] `removeEntity()` 
- [ ] `entity(vector<Entity>)`

which are mostly clones from each other. While this is ok for the front-facing objects It is at least ugly in the actual implementations.

jgrewe:
has been greatly reduced with pull request #679

PR #620, opened 2016-08-02
Inverse search of DataArrays, Tags, etc.. from a section

jgrewe:
This pr introduces the long missing convenience methods for inverse search from Sections to referring entities(issue #551). Unfortunately, it is not possible to implement the functionality using templates due to circular includes. 

This relates to issue #560.

### DIFF
--- a/src/Section.cpp
+++ b/src/Section.cpp
@@ -341,8 +341,17 @@ std::vector<nix::DataArray> Section::referringDataArrays() const {
     std::vector<nix::DataArray> arrays;
     nix::File f = backend()->parentFile();
     for (auto b : f.blocks()) {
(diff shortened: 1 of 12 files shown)


src/Section.cpp
```diff
@@ -332,3 +335,89 @@ Section Section::getSection(ndsize_t index) const {
...
+std::vector<nix::Source> Section::referringSources(const Block &b) const {
+    std::vector<nix::Source> srcs;
+    if (b) {
+        srcs = b.findSources(nix::util::MetadataFilter<nix::Source>(id()));
+    }
+    return srcs;
+}
+
+
+std::vector<nix::Block> Section::   referringBlocks() const {
```
gicmo: Accidental whitespace?

PR #623, opened 2016-08-04
Inverse search of DataArrays, Tags, etc.. from a Source

jgrewe:
Implementation is not optimal since we do not have a direct access to the block in which the source resides. Thus, the methods search across all Blocks...

### DIFF
--- a/backend/fs/EntityWithSourcesFS.cpp
+++ b/backend/fs/EntityWithSourcesFS.cpp
@@ -51,7 +51,11 @@ ndsize_t EntityWithSourcesFS::sourceCount() const {
 
 
 bool EntityWithSourcesFS::hasSource(const std::string &id) const {
-    return !sources_dir.findByNameOrAttribute("entity_id", id)->empty();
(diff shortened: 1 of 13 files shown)


backend/fs/EntityWithSourcesFS.cpp
```diff
@@ -51,7 +51,11 @@ ndsize_t EntityWithSourcesFS::sourceCount() const {
 
 
 bool EntityWithSourcesFS::hasSource(const std::string &id) const {
-    return !sources_dir.findByNameOrAttribute("entity_id", id)->empty();
+    boost::optional<bfs::path> p = sources_dir.findByNameOrAttribute("entity_id", id);
+    if (p)
+        return true;
+    else
+        return false;
```
gicmo: This looks funny, but `return p;` won't work for contextual bool conversions IIRC . `return !!p;` should, but I am not sure that looks better. Maybe 'return p ? true : false;' or `return p.is_initialized();`

jgrewe: agreed! not nice. return p; won't work because "their is no known conversion to bool"... 
is_initialized has been deprecated

PR #631, opened 2016-09-29
Dimensions: safe addition, remove redundant checks

gicmo:


jgrewe:
Grml, seems, that gcc does not have a __has_builtin macro ...

src/Dimensions.cpp
```diff
@@ -376,15 +373,19 @@ size_t RangeDimension::indexOf(const double position) const {
 
 vector<double> RangeDimension::axis(const ndsize_t count, const ndsize_t startIndex) const {
 
-    size_t cnt = check::fits_in_size_t(count, "Axis count exceeds memory (size larger than current system supports)");
-    size_t idx = check::fits_in_size_t(startIndex, "Axis start index exceeds memory (size larger than current system supports)");
```
gicmo: That check needs to stay, but moved down I guess.

include/nix/Platform.hpp
```diff
@@ -43,4 +43,8 @@
 #define NIX_SRC_FUNC  __PRETTY_FUNCTION__
 #else
 #define NIX_SRC_FUNC __FUNCSIG__
-#endif
\ No newline at end of file
+#endif
+
+#ifndef __has_builtin
+#define __has_builtin(x) 0
+#endif
```
jgrewe: that's quite a different solution from what I had in mind yesterday night. Good, that you did it :)

Issue #635, opened 2016-10-19
Change boolean representation


achilleas-k:
In order to solve g-node/nixpy#212, we will have to change the way NIX stores boolean data. NIX should store this data in the same way as h5py to be compatible with NIXPY. For backwards compatibility, NIX should maintain the ability to read bools as bitfields, but any new data should be written in the h5py compatible way.

gicmo:
This should be fixed.

PR #637, opened 2016-10-26
New boolean representation

achilleas-k:
Booleans are now stored as enums with members "FALSE" and "TRUE".

Fixes issue #635.


backend/hdf5/h5x/H5DataType.cpp
```diff
@@ -281,6 +281,12 @@ data_type_from_h5(H5T_class_t vclass, size_t vsize, H5T_sign_t vsign)
     return DataType::Nothing;
 }
 
+void check_bool_enum(h5x::DataType type) {
+    assert(type.member_count() == 2);
```
gicmo: `assert` statements can be turned into NOOPs with the right define (NDEBUG), which is not a good idea. Throwing an exception, or returning `DataType::Nothing`.

gicmo: On a second though, you should be able to use `boolfiletype.enum_equal(type)` for the check.

achilleas-k: I used `assert` since it's also used in `data_type_from_h5` when the the `vclass` is unknown. It's also used when the member count in the compound data type is wrong.

Should those cases be throwing exceptions as well?

gicmo: `Assert` is to find programming errors. In general, if we didn't fuck up, then there should never be a case where any of the asserts are it. If for whatever reason that happens then in `data_type_from_h5` we return a `DataType::Nothing`, which is one of the sensible things to do. Throwing an exception would also work, but I guess the user should never actually see it.
In the bool case if the asserts are "not working" then we would accept something not a bool as a bool which we really shouldn't do.

PR #642, opened 2016-10-30
[example data] with old style boolean value

jgrewe:
for checks of downward compatibility and maybe the migration tool


gicmo:
Nice! Thanks a lot. One thing, is `docs/data` the right place? should it not be in `test/hdf5/data` or something like this?

jgrewe:
will create a different file and move it

PR #643, opened 2016-10-30
Appveyor configuration file

achilleas-k:
Exported `appveyor.yml` from web configuration for easier editing in repository and testing in forks.


CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ cmake_minimum_required (VERSION 2.8.10)
 project (nix C CXX)
 
 set(VERSION_MAJOR 1)
-set(VERSION_MINOR 1)
+set(VERSION_MINOR 2)
 set(VERSION_PATCH 0)
```
jgrewe: good catch, was looking in the appveyor settings for the reason why they are still called 1.0.0

gicmo: On master we should in the future probably bump the versions immediately after the release.

Issue #647, opened 2016-11-04
add function that returns the library version


jgrewe:

achilleas-k:
Solved via PR #649.

PR #649, opened 2016-11-05
add static method for library version to File

jgrewe:
version header is created from the cmake settings. should solve issue #647

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@ cmake_minimum_required (VERSION 2.8.10)
 project (nix C CXX)
 
 set(VERSION_MAJOR 1)
-set(VERSION_MINOR 2)
+set(VERSION_MINOR 3)
 set(VERSION_PATCH 0)
 
 set(VERSION_ABI   1)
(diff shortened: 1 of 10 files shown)


CMakeLists.txt
```diff
@@ -110,6 +109,19 @@ foreach(backend ${backends})
...
+### AUTO-GENERATED VERSION HEADER
+
+configure_file (
+  "${PROJECT_SOURCE_DIR}/version.h.in"
+  "include/nix/libversion.hpp"
+  )
+ 
+# add the binary tree to the search path for include files
+# so that we will find version.h
+include_directories("${PROJECT_BINARY_DIR}/include/nix")
```
gicmo: What if we are change that to `include_directories("${PROJECT_BINARY_DIR}/include/")` and then later on include `<nix/libversion.hpp>`.

include/nix/File.hpp
```diff
@@ -16,6 +16,7 @@
 #include <nix/Platform.hpp>
 
 #include <nix/valid/validate.hpp>
+#include "libversion.hpp"
```
gicmo: Should we move this include to `nix/Version.hpp`?

jgrewe: would you shift the version method to Version.hpp as well?

include/nix/Version.hpp
```diff
@@ -2,12 +2,20 @@
...
 #define NIX_VERSION_HPP
 
 namespace nix {
-
+/**
+ * @brief Returns the version of the library.
+ * 
+ * @return the version as a vector<int> in the order major, minor, patch
+ */
+std::vector<int> apiVersion();
```
gicmo: missing `NIXAPI` ;)

jgrewe: do not merge, I will rebase once I have that fixed

PR #651, opened 2016-11-11
Read bitfields as booleans

achilleas-k:
Registers a converter in HDF5 for reading `H5T_STD_B8LE`, the old boolean type, found in files.
Notice that the converter doesn't really do anything. I suspect it might be implicitly casting data to the destination type though I haven't found a definitive description in the HDF5 documentation that explains this.

If you think it's better that we define an explicit conversion path, for instance if the current method is simply exploiting behaviour that is not guaranteed, I'll do it and push to this PR.

### DIFF
--- a/backend/hdf5/h5x/H5DataType.cpp
+++ b/backend/hdf5/h5x/H5DataType.cpp
@@ -9,16 +9,12 @@
 
 #include "H5DataType.hpp"
 
-#include <stdexcept>
-#include <iostream>
-#include <cassert>
-#include <string>
 #include <memory>
-#include <cstdlib>
 
 namespace nix {
 namespace hdf5 {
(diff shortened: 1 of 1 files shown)


backend/hdf5/h5x/H5DataType.cpp
```diff
@@ -9,17 +9,14 @@
 
 #include "H5DataType.hpp"
 
-#include <stdexcept>
-#include <iostream>
-#include <cassert>
-#include <string>
 #include <memory>
-#include <cstdlib>
+#include <malloc.h>
```
jgrewe: Do we need the malloc include? Under linux and OS X it works without? If windows needs it, I am afraid that we need to find a way to only include it if os == win

PR #653, opened 2016-11-19
increase version number to 1.3.1

jgrewe:
I would like to publish the cmake fix as 1.3.1 so we need to increase the patch version...


gicmo:
In the future we should do version changes directly after the release (and no need for a pull request, I guess) :).

PR #662, opened 2017-02-06
Visual studio compile fixes

The-E:
While looking through the nix source, I noticed a few areas where nix wasn't compatible with MS' Visual C compiler. This PR fixes those, but there is still more to do in this area; there are a few things I would like to do to make setting up the dependencies used easier.

### DIFF
--- a/backend/fs/DataArrayFS.cpp
+++ b/backend/fs/DataArrayFS.cpp
@@ -12,16 +12,13 @@
 #include "hdf5/h5x/H5DataSet.hpp" // FIXME
 #include "DimensionFS.hpp"
 
-namespace bfs = boost::filesystem;
-
 namespace nix {
 namespace file {
 
-
 DataArrayFS::DataArrayFS(const std::shared_ptr<base::IFile> &file, const std::shared_ptr<base::IBlock> &block,
-                         const std::string &loc)
+                         const bfs::path &loc)
     : EntityWithSourcesFS(file, block, loc),
-      dimensions(loc + bfs::path::preferred_separator + "dimensions", file->fileMode()) {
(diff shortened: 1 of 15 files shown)


jgrewe:
Hi The-E, many thanks for the pull request. We appreciate your help in the windows world a lot! 
Under Linux (boost 1.54, the current boost version on trusty, i.e. the travis builds) the use of filesystem::path::append breaks. So far we require boost >= 1.49, I would rather keep this dependency.

backend/fs/DataArrayFS.cpp
```diff
@@ -12,30 +12,41 @@
...
 namespace nix {
 namespace file {
 
-
 DataArrayFS::DataArrayFS(const std::shared_ptr<base::IFile> &file, const std::shared_ptr<base::IBlock> &block,
-                         const std::string &loc)
+                         const bfs::path &loc)
     : EntityWithSourcesFS(file, block, loc),
-      dimensions(loc + bfs::path::preferred_separator + "dimensions", file->fileMode()) {
+      dimensions(bfs::path(loc).append("/dimensions"), file->fileMode()) {
```
jgrewe: Does this work on the windows side (using the '/' as a separator)? Good to know

The-E: boost::filesystem automatically changes any path separators into the native format.

achilleas-k:
Hey @The-E, thanks for the PR.
As @jgrewe mentioned, we want to keep compatibility with boost 1.49 for NIX on all platforms, even though we don't explicitly test for it on macOS and Windows.

Could you revert the changes that broke compatibility with 1.49? I think it's the path append functions that are the issue, but it's not clear from the Travis logs.

jgrewe:
closing this one, opened a replacement that also fixes the linux builds

PR #664, opened 2017-02-18
Replacement for apparently stalled pr from The-E

jgrewe:
see #662 for more info


gicmo:
i am still unhappy about the huge commit message. A `rebase -i` with some rewording would be nice, although if it is fine for you as-is I am not opposing it too much ;)

Issue #650, opened 2016-11-11
Tag/Multitag hasFeature(name) and getFeature(name) do not work properly


jgrewe:

jgrewe:
Actually this is no bug but a "non-feature", i.e. the method is only supposed to take an id. In case one passes a name, it does not complain but also does not return a thing. One wonders, if we should add searching by name.

achilleas-k:
It would be nice to support retrieval by name. I think it someone might expect that it would work since retrieving items by name in addition to ID is common throughout the library.

PR #669, opened 2017-04-09
Add function overloads that allow accessing reference and feature data by name ...

jgrewe:
get- and hasFeature/Reference also accepts the name or id of the searched entity (issue #650). Heuristic first searches among the id of the features for direct hits otherwise it the names and ids of the linked dataArrays are checked.

### DIFF
--- a/backend/fs/BaseTagFS.cpp
+++ b/backend/fs/BaseTagFS.cpp
@@ -155,7 +155,6 @@ void BaseTagFS::references(const std::vector<DataArray> &refs_new) {
 
 bool BaseTagFS::hasFeature(const std::string &name_or_id) const {
     return getFeature(name_or_id) != nullptr;
-    //!feature_group.findByNameOrAttribute("entity_id", name_or_id)->empty();
 }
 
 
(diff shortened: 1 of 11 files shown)


backend/fs/BaseTagFS.cpp
```diff
@@ -154,7 +154,8 @@ void BaseTagFS::references(const std::vector<DataArray> &refs_new) {
 //--------------------------------------------------
 
 bool BaseTagFS::hasFeature(const std::string &name_or_id) const {
-    return !feature_group.findByNameOrAttribute("entity_id", name_or_id)->empty();
+    return getFeature(name_or_id) != nullptr;
+    //!feature_group.findByNameOrAttribute("entity_id", name_or_id)->empty();
```
gicmo: Just get rid of this, we have version control for that ;)

src/MultiTag.cpp
```diff
@@ -104,6 +104,16 @@ DataView MultiTag::retrieveData(size_t position_index, size_t reference_index) c
 }
 
 
+DataView MultiTag::retrieveData(size_t position_index, const std::string &name_or_id) const {
+    nix::DataArray array = backend()->getReference(name_or_id);
+    if (array) {
+        return util::retrieveData(*this, position_index, array);
+    } else {
+        throw nix::OutOfBounds("There is no DataArray with the specified name or id!", 0);
```
gicmo: Not sure OutOfBounds is the right exception here. Maybe something derived from argument_error (or that itself)?

jgrewe: will be fixed...

Issue #672, opened 2017-04-25
Unexpected (incorrect) behaviour when adding multiple objects to a group


achilleas-k:
## Demonstration of problem

Consider the following example program:
```c++
#include <nix.hpp>


int main(){
    nix::File file = nix::File::open("append_bug.nix", nix::FileMode::Overwrite);
    nix::Block bone = file.createBlock("One", "test");
    nix::Block btwo = file.createBlock("Two", "test");
    nix::DataArray done = bone.createDataArray("DATA", "test", nix::DataType::Double, {1});
    nix::DataArray dtwo = btwo.createDataArray("DATA", "test", nix::DataType::Double, {1});  // same name

    nix::Group grp = bone.createGroup("grp", "test");

//    grp.addDataArray(dtwo);  // Uncomment to throw error

    grp.dataArrays({dtwo});  // this should throw error but succeeds

    std::cout << "dtwo name and id:     " << dtwo.name() << ", " << dtwo.id() << std::endl;
    std::cout << "appended name and id: " << grp.dataArrays()[0].name() << ", " << grp.dataArrays()[0].id() << std::endl;

    return 0;
}
```

The line where `dtwo` is added to `grp` via `dataArrays(<vector>)` succeeds even though it should fail. However, the object that gets appended isn't the one given to the method.

### Output
Sample output of the program:
```
dtwo name and id:     DATA, 0f5a0c71-c37e-40da-83a3-8c27ca923a86
appended name and id: DATA, 7a6fe116-5dd1-4038-97ef-a573c9a43789
```

## Description of behaviour
The `dataArrays()` method (https://github.com/G-Node/nix/blob/master/backend/hdf5/GroupHDF5.cpp#L108) extracts the names from the provided vector of DataArrays. It uses the names  to check if the parent block contains the provided DataArrays. If it finds the DataArrays in question, it extracts their ID and passes each ID on to the `addDataArray` function.

The problem occurs when a DataArray passed to the `dataArrays()` method has the same name as a DataArray in the parent block, but is not the same object (different ID and presumably different data). The valid object gets appended, i.e., the DataArray that exists on the parent object, but this is not the DataArray that the user provided.

## Solution
Once an object is obtained from the parent Block, compare the IDs as well.

jgrewe:
the other solution to the problem would be to disallow adding data_arrays from different blocks... This is, what we do in MultiTags for example.

```
if (!block()->hasDataArray(name_or_id))
[...]

PR #674, opened 2017-05-08
Fixing bug #672

jgrewe:
all(?!) vector setters verify entity identity by name and id

change behaviour of has-, add-, removeEntity methods to use the name instead of the id

### DIFF
--- a/backend/fs/BaseTagFS.cpp
+++ b/backend/fs/BaseTagFS.cpp
@@ -110,42 +110,32 @@ bool BaseTagFS::removeReference(const std::string &name_or_id) {
 
 
 void BaseTagFS::references(const std::vector<DataArray> &refs_new) {
-    // extract vectors of names from vectors of new & old references
-    std::vector<std::string> names_new(refs_new.size());
-    transform(refs_new.begin(), refs_new.end(), names_new.begin(), util::toName<DataArray>);
-
-    size_t count = nix::check::fits_in_size_t(referenceCount(), "referenceCount() failed! count > than size_t!");
-    std::vector<DataArray> refs_old(count);
-    for (size_t i = 0; i < refs_old.size(); i++){
-        refs_old[i] = getReference(i);
+    auto cmp = [](const DataArray &a, const DataArray& b) { return a.name() < b.name(); };
+    std::vector<DataArray> new_arrays(refs_new);
+    size_t ref_count = nix::check::fits_in_size_t(referenceCount(), "referenceCount() failed; count > size_t.");
+    std::vector<DataArray> old_arrays(ref_count);
+    for (size_t i = 0; i < old_arrays.size(); i++) {
+        old_arrays[i] = getReference(i);
     }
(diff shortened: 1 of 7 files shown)


backend/fs/BaseTagFS.cpp
```diff
@@ -110,42 +110,32 @@ bool BaseTagFS::removeReference(const std::string &name_or_id) {
...
-    std::set_difference(names_new.begin(), names_new.end(), names_old.begin(), names_old.end(),
-                        std::inserter(names_add, names_add.begin()));
-    std::set_difference(names_old.begin(), names_old.end(), names_new.begin(), names_new.end(),
-                        std::inserter(names_rem, names_rem.begin()));
-
-    // check if all new references exist & add sources
     auto blck = std::dynamic_pointer_cast<BlockFS>(block());
-    for (auto name : names_add) {
-        if (!blck->hasDataArray(name))
+    for (auto da : add) {
```
gicmo: Could and probably should here (and in all following places) be: `const auto &da`, because we are copying the `DataArray` object itself otherwise.

backend/fs/BaseTagFS.cpp
```diff
@@ -110,42 +110,32 @@ bool BaseTagFS::removeReference(const std::string &name_or_id) {
...
+        DataArray a = blck->getDataArray(da.name());
+        if (!a || a.id() != da.id())
             throw std::runtime_error("One or more data arrays do not exist in this block!");
-        addReference(blck->getDataArray(name)->id());
+        addReference(a.id());
```
gicmo: Not new, so not a blocker for this change, but `addReference` does a lot of work again that we just did.

backend/fs/BaseTagFS.cpp
```diff
@@ -110,42 +110,32 @@ bool BaseTagFS::removeReference(const std::string &name_or_id) {
...
-
-    size_t count = nix::check::fits_in_size_t(referenceCount(), "referenceCount() failed! count > than size_t!");
-    std::vector<DataArray> refs_old(count);
-    for (size_t i = 0; i < refs_old.size(); i++){
-        refs_old[i] = getReference(i);
+    auto cmp = [](const DataArray &a, const DataArray& b) { return a.name() < b.name(); };
+    std::vector<DataArray> new_arrays(refs_new);
+    size_t ref_count = nix::check::fits_in_size_t(referenceCount(), "referenceCount() failed; count > size_t.");
+    std::vector<DataArray> old_arrays(ref_count);
+    for (size_t i = 0; i < old_arrays.size(); i++) {
```
gicmo: Look, I can save one line but make it more complicated: ;)
```cpp
size_t i = 0;
std::generate(old_arrays.begin(), old_arrays.end(), [&i, this]{ return getReference(i++); });
```

backend/fs/BaseTagFS.cpp
jgrewe: not sure how to avoid this instead of avoiding the call altogether

Issue #673, opened 2017-04-25
Appending objects to groups should not use ID getters


achilleas-k:
The append methods for the Group object should not use IDs for retrieving objects from the parent block. Using the ID to check if the object exists and for retrieving it for append causes the operation to become slower as the number of objects increases.

Ideally, if an object is provided to the `add` method, its name should be used to retrieve the object from the parent block and the ID should be compared to make sure it is the same object. This is similar to how `dataArrays()` appends multiple items, but we should also include an ID comparison (see #672).

jgrewe:
solved via #679

PR #679, opened 2017-05-28
Generic entity access methods for Block

gicmo:
Instead of having specialised {has,get,count,remove}XYZ methods defined on `IBlock`,
we now have  generic methods; this removes code duplication in the backend.
This is also step one to generalise the bulk replacement of entities.
Additionally a new Identity class is used for these methods that should solve issue #673.
The current PR does it for `Block` and the `hdf5` backend. Will break the `filesystem`
backend.

### DIFF
--- a/backend/fs/BlockFS.hpp
+++ b/backend/fs/BlockFS.hpp
@@ -33,6 +33,10 @@ private:
 
     void createSubFolders(const std::shared_ptr<base::IFile> &file);
 
+    // Helper methods for generic entity related methods below
+    boost::optional<Directory> groupForObjectType(ObjectType ot) const;
+
+    boost::optional<boost::filesystem::path> findEntityGroup(const nix::Identity &ident) const;
 public:
 
     /**
@@ -67,6 +71,23 @@ public:
     BlockFS(const std::shared_ptr<base::IFile> &file, const std::string &loc, const std::string &id, const std::string &type, const std::string &name, time_t time);
 
     //--------------------------------------------------
+    // Generic entity methods
+    //--------------------------------------------------
+    std::string resolveEntityId(const nix::Identity &ident) const;
+
+    bool hasEntity(const nix::Identity &ident) const;
+
+    std::shared_ptr<base::IEntity> getEntity(const nix::Identity &ident) const;
+
+    std::shared_ptr<base::IEntity> getEntity(ObjectType type, ndsize_t index) const;
+
+    ndsize_t entityCount(ObjectType type) const;
+
+    bool removeEntity(const nix::Identity &ident);
+
+    void addEntity(const nix::Identity &ident);
+
+    //--------------------------------------------------
     // Methods concerning sources
     //--------------------------------------------------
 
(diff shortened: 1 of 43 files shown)


gicmo:
Maybe you guys could have a brief glance over it, just in case we are doing something obviously stupid.

cgars:
A i see the fs stuff came late to the party... Sry for the buzz

backend/fs/BlockFS.hpp
```diff
@@ -67,6 +71,23 @@ class BlockFS : virtual public base::IBlock, public EntityWithMetadataFS,
     BlockFS(const std::shared_ptr<base::IFile> &file, const std::string &loc, const std::string &id, const std::string &type, const std::string &name, time_t time);
 
     //--------------------------------------------------
+    // Generic entity methods
+    //--------------------------------------------------
+    std::string resolveEntityId(const nix::Identity &ident) const;
```
cgars: [npick] the Identities are  sometimes called ident and sometime id...

gicmo: ah, good catch, thanks

gicmo:
Well, as long as the front end methods keep `name_or_id` we will have the ambiguity. I wasn't aware that we wanna get rid of them (they are quite convenient for API users I guess). And you are right currently the behavior of the code is to work with happily with incomplete `Identity`s, i.e. if we have incomplete information (i.e. `name_or_id`) we set only one of the fields. Maybe we should change the frontend objects to have functions with dedicated id and name parameters but that would be a bigger API change. The current (already quite big) changeset is only changing internal things; but it would certainly pave the way for that bigger API change.

achilleas-k:
That's fair. I agree on not changing the API for now. I was mostly concerned about the behaviour of the Identity class. The API and internals could be the same with the small difference that the Identity is always a name *and* ID, so when a method requires creating an Identity object but only has access to a name, it would be the responsibility of the method to fetch the ID first and then create an Identity, rather than creating an Identity with name only and using that to fetch the ID.

backend/fs/BlockFS.hpp
jgrewe: true and not nice... but backend, is that tolerable for now?

gicmo: Gimme a moment, I will fix it in a bit.

jgrewe:
From the user's perspective the name is the most convenient and natural thing to do. In relacs we often do not have the concrete Entity at hand (and hence do not have the id) but there is a rule how unique names are generated and I can look if, e.g. a Tag, with such a name already exists. Requiring name and id would mean to have access to the actual entity, and then I probably do not need the method anymore. I see your point but do not see, how we can avoid this entirely without a loss in usability.

achilleas-k:
But the Identity class isn't user-facing, is it?

Let me clarify my position: I'm mostly thinking about future development falling into the same trap of identifying objects by `name` or `id`, but not both. I saw the introduction of this new class and methods as a way to _enforce_ using name and id (internally) for identification of objects. Allowing Identity objects to possess only one of the two doesn't enforce this. My thinking is that it would be nice if we (not the user) can be certain that any Identity instance refers to an object and is 100% unambiguous, because it is guaranteed to have a name and an id.

gicmo:
I guess if we had the `resolveNameOrId(const std::string &name_or_id)` and return an `Identity` we could use in all user facing methods that get `name_or_id` and then internally we would only have unambiguous `Identity`s. That is what @achilleas-k wants, right?

achilleas-k:
Exactly

PR #681, opened 2017-06-06
[Test] Failing test for read only group creation

achilleas-k:
New test for retrieving information (count) for groups that do not exist and the file is open Read Only.

Test fails when requesting the number of items in a container group that
doesn't exit and the file is in open RO.


achilleas-k:
Included in #682. Closing.

PR #682, opened 2017-06-06
CI test fixes, Group creation in RO mode fixes 

gicmo:
Tested the failed didn't make the CI fail (#681). Also PR #679 introduced a regression for entity access in ReadOnly mode; we also lacked a test for that.


achilleas-k:
Now we need to figure out why Appveyor tests are succeeding.

gicmo:
@achilleas-k Appveyor is not building the FS backend, and that is the only backend that was/is failing. Good question why the HDF5 one is not.

achilleas-k:
So like Travis, the test in Appveyor does [fail](https://ci.appveyor.com/project/achilleas-k/nix-n65s7/build/1.2.112/job/cikhsdr3q7sfoyk7#L930), but the build succeeds.

achilleas-k:
I think the problem is with the [build_script](https://github.com/G-Node/nix/blob/master/appveyor.yml#L22) block. It's a single command block, which means that even if the `ctest` line fails, the next command will execute and if the last one succeeds, the whole block succeeds.

We need to either separate each command on its own line or move the tests to the `test_script` block.

achilleas-k:
So are we good to merge this and then open the PR to fix the issues?

Issue #683, opened 2017-06-07
Appending multiple sources simultaneously does not respect order


achilleas-k:
While working on the Matlab bindings, @mpsonntag found the following issue.

When appending multiple sources to an EntityWithSources (tested with Group), they are either not appended in order, or the retrieval does not happen in order.

The following code demonstrates the issue:
```c++
#include <nix.hpp>

int main(){
    nix::File file = nix::File::open("extendorder.h5", nix::FileMode::Overwrite);
    nix::Block b = file.createBlock("blk", "blk");
    nix::Group g = b.createGroup("group", "grp");

    for (int idx = 0; idx < 10; idx++) {
        b.createSource(std::to_string(idx), "src");
    }
    
    g.sources(b.sources());
    for (auto src : g.sources()) {
        std::cout << src.name() << std::endl;
    }
    file.close();

    return 0;
}
```
(output is random).

Replacing `g.sources(b.sources())` with a loop that appends sources in order works as expected however.

jgrewe:
actually it works as specified :) 

The order changes due to the linking mechanism. The link is done using the id, not the name. If you print out the id it is in perfect order.

achilleas-k:
Interesting.
Is this the intended behaviour? Should we be sorting by supplied order (as it appears in the vector)?

jgrewe:
ok ok, I have to correct myself. When we create the group we specify that the oder should be by creation oder.
```HErr res = H5Pset_link_creation_order(gcpl.h5id(),  H5P_CRT_ORDER_TRACKED | H5P_CRT_ORDER_INDEXED);```

In the vector setter we order the passed entities by name, create the differences with the already present once,  remove  and add according to name.

The other thing I wonder: Do we really want to go long ends to maintain the order (any order)?

jgrewe:
should we leave this issue open until we implement the util function for custom ordering?

gicmo:
@jgrewe I'd say we open a new one for the util, and close this.

achilleas-k:
Agreed

PR #686, opened 2017-06-08
Ordering of sources in EntitiesWithSources

jgrewe:
solves issue #683

### DIFF
--- a/backend/fs/EntityWithSourcesFS.cpp
+++ b/backend/fs/EntityWithSourcesFS.cpp
@@ -43,7 +43,7 @@ EntityWithSourcesFS::EntityWithSourcesFS(const std::shared_ptr<base::IFile> &fil
 EntityWithSourcesFS::EntityWithSourcesFS(const std::shared_ptr<base::IFile> &file, const std::shared_ptr<base::IBlock> &block,
 										 const std::string &loc, const std::string &id, const std::string &type,
 										 const std::string &name)
-	: EntityWithSourcesFS(file, block, bfs::path(loc), id, type, name) 
+	: EntityWithSourcesFS(file, block, bfs::path(loc), id, type, name)
 {
 }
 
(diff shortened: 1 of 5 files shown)


achilleas-k:
Wrote a test for it (sources only, HDF5 only) here, if you feel like pulling it in:
https://github.com/achilleas-k/nix/commit/abd8ba3d46c041892fe11034186e425de785e91c

jgrewe:
@achilleas-k Thanks for the test!

test/BaseTestEntityWithSources.cpp
```diff
@@ -82,4 +82,30 @@ void BaseTestEntityWithSources::testSourceVectorSetter() {
...
\ No newline at end of file
+}
+
+void BaseTestEntityWithSources::testSourceOrder() {
+    std::vector<std::string> names = { "source_a", "source_b", "source_c", "source_d", "source_e" };
+    std::vector<Source> sources;
+    Group append_group = block.createGroup("ag", "test");
+    Group set_group = block.createGroup("sg", "test");
+
+    for (auto it = names.begin(); it != names.end(); it++) {
```
gicmo: For future reference, "range based for loops" are nicer; i.e. `for (const auto &name : names) { ... }`.
I assume this was c&p'ed from code above, so one could make an argument about consistency, even though the iterator based approach is just so C++03 ;). OTOH you are using range based loop below, sooo ;) .. Anyway, I'll allow it ;)

test/BaseTestEntityWithSources.cpp
```diff
@@ -82,4 +82,30 @@ void BaseTestEntityWithSources::testSourceVectorSetter() {
...
+        sources.push_back(block.createSource(*it, "channel"));
+    }
+
+    // append each source in order
+    for (auto src : block.sources()) {
```
gicmo: `block.sources()` returns a `vector<Source>`, so src is of type `Source`, which is ok (we are allocating a Source on the stack and copying a `shared_ptr<ISource>` (via the copy constructor(s)); slightly better, `const auto &src`, now unsurprisingly `src` is of type `const Source &` and we are basically just passing a pointer around. Totally ok to not do in a test, also allowed ;)

test/BaseTestEntityWithSources.cpp
```diff
@@ -82,4 +82,30 @@ void BaseTestEntityWithSources::testSourceVectorSetter() {
...
+        append_group.addSource(src);
+    }
+
+    // use the vector setter
+    set_group.sources(sources);
+
+    for (size_t idx = 0; idx < sources.size(); ++idx) {
+        CPPUNIT_ASSERT(sources[idx].name() == append_group.getSource(idx).name());
```
gicmo: `CPPUNIT_ASSERT_EQUAL(expected, actual)` gives nicer errors in case of broken tests. Still ok.

backend/fs/EntityWithSourcesFS.cpp
```diff
@@ -43,7 +43,7 @@ EntityWithSourcesFS::EntityWithSourcesFS(const std::shared_ptr<base::IFile> &fil
 EntityWithSourcesFS::EntityWithSourcesFS(const std::shared_ptr<base::IFile> &file, const std::shared_ptr<base::IBlock> &block,
 										 const std::string &loc, const std::string &id, const std::string &type,
 										 const std::string &name)
-	: EntityWithSourcesFS(file, block, bfs::path(loc), id, type, name) 
+	: EntityWithSourcesFS(file, block, bfs::path(loc), id, type, name)
```
gicmo: Meh, whitespace changes ;)

backend/fs/EntityWithSourcesFS.cpp
```diff
@@ -97,38 +97,35 @@ std::shared_ptr<base::ISource> EntityWithSourcesFS::getSource(const size_t index
...
+    std::vector<Source> add;
+    std::vector<Source> rem;
+
+    std::set_difference(new_arrays.begin(), new_arrays.end(), old_arrays.begin(),
+                        old_arrays.end(), std::inserter(add, add.begin()), cmp);
+    std::set_difference(old_arrays.begin(), old_arrays.end(), new_arrays.begin(),
+                        new_arrays.end(), std::inserter(rem, rem.begin()), cmp);
+
+    for (const auto &e : add) {
+        addSource(e.id());
```
gicmo: Is there are a reason we can't just pass the whole entity (i.e. `e`); same below.

test/BaseTestEntityWithSources.cpp
```diff
@@ -82,4 +82,30 @@ void BaseTestEntityWithSources::testSourceVectorSetter() {
     std::vector<Source> deleter;
     da.sources(deleter);
     CPPUNIT_ASSERT(da.sourceCount() == 0);
-}
\ No newline at end of file
+}
+
+void BaseTestEntityWithSources::testSourceOrder() {
+    std::vector<std::string> names = { "source_a", "source_b", "source_c", "source_d", "source_e" };
```
gicmo: One thing: maybe we should randomise that order (maybe even in code, with `std::random_shuffle`). I am actually not entirely convinced that this wouldn't break the existing code. But it is late and I might be wrong.

achilleas-k: I agree it would be a better test. For instance, I can imagine a situation where the order isn't tracked but the Sources are returned in alphabetical order by name (say, in a future FS backend that links by name).
Randomising is a good way of ensuring tracking works as expected.

jgrewe: @gicmo is right, test will fail on random order. I  fixed the particular issue with the naming order but this does not generalize... :(
At this point we need a poll, methinks. What order should it be? Creation order? Name order? Order of things in the passed vector?

The easiest, and maybe best solution is to just take the order of the passed vector. This would mean that we simplify the whole function by removing all the old junk and simply setting the new stuff. We might even gain some performance in the hdf and fs backend (since we have quite a bit of file access there) but may loose some in a memory-backend...
 

If we want to have the creation order: is it the order in which we create the links to the entities or the time, the entities themselves have been created? Our created at will not be sufficient for such decisions and from the frontend we do not have access  to the real one...

gicmo:
So, I am afraid I was right: I made a quick patch to randomise the list of names in the sources vector (see and maybe pull from here: https://github.com/gicmo/nix/tree/pr/686) and this results in:
```
Test name: TestEntityWithSourcesHDF5::testSourceOrder
assertion failed
- Expression: sources[idx].name() == set_group.getSource(idx).name()
```
So if I read issue #683 correctly then this patch does not actually fix that issue (hence my confusion looking at the code yesterday). To properly fix it we have to retain the order of the entities to be added, i.e. the `add` vector needs to be sorted to match the incoming `sources` vector. I guess it is obvious but could be done with a combination of `std::sort` with a lambda using `std::find` on the original `sources` vector. Should be quickly implemented but maybe not super efficient (although the obvious LUT approach might not be faster for the few items we have in the vectors normally).

test/BaseTestEntityWithSources.cpp
jgrewe: what about this one:

````
void EntityWithSourcesHDF5::sources(const std::vector<Source> &sources) {
    while (sourceCount() > 0) {
        removeSource(getSource(0)->id());
    }
    for (const auto &src : sources) {
        if (block()->hasEntity(src) ) {
            addSource(src.id());
        }
    }
}
````

**Comment by achilleas-k** [comment]
I think I need to rewind a bit because I lost track of the general idea here.

*Correct me if I'm wrong*: The vector setter (I'm calling it that now because that's what the corresponding test calls it) replaces all referenced sources. It does this by adding Sources that are in `new_arrays` but not in `old_arrays` and removing sources that are in `old_arrays` but not in `new_arrays`.

If this is the case, preserving order becomes a bit tricky, right? What's the logical result of the following (simplified) code snippet?

```C++
group.sources({A, B, C});
assert(A == group.sources(0));  // at least it should

group.sources({A, D, B, E, C});
assert(D == group.sources(1));  // Is it?
assert(B == group.sources(1));  // or maybe this?
```

Which of the two last asserts fails? (or I guess, which one *should* fail?)

I'm not entirely sure what the behaviour of this is right now, but I would expect D to be in the second position there.

jgrewe:
@achilleas-k you are right, in the original version we add only those that are not already there and remove those that have been there but are not in the argument vector. 

So far, it has been sorted by the ID, thus, we cannot tell in your example. The setter in this pr would sort according to the name of the entity and then 
````
assert(B == group.sources(1));
````
would be working. 

The simplified version above would now simply take the order given by the input vector and
````
assert(D == group.sources(1));
````
would be correct.
I tend to this solution. It is rather straight forward and may even be faster than all the io traffic for checking who is there...

jgrewe:
will address this issue later today

PR #688, opened 2017-06-10
again the sources

jgrewe:
replacement for PR #686 
also simplifies the generic vector setter in Group and the DataArray setters in BaseTagFS/HDF5


achilleas-k:
Minor comment: Best PR title ever?

PR #692, opened 2017-08-02
Section find does not include self in result set 

### LINKED ISSUE #690: 'Section.findSections' returns its own Section in the resultset
When calling `Section.findSections`, the result set of subsections also contains itself (the section from which the function was called) if the used filter applies.
The question arose whether the expected behavior is rather to exclude itself from the search result.

jgrewe:
fixes #690

### DIFF
--- a/src/Section.cpp
+++ b/src/Section.cpp
@@ -89,7 +89,7 @@ void addChildrenIfNotMaxDepth(std::tuple<Section, size_t>& current,
     if (std::get<1>(current) < max_depth) {
         size_t next_depth = std::get<1>(current) + 1;
         for (const auto &s : std::get<0>(current).sections()) {
-            todo.push_back(std::make_tuple(s, next_depth));
+            todo.emplace_back(s, next_depth);
         }
     }
 }
--- a/src/Section.cpp
+++ b/src/Section.cpp
@@ -83,32 +83,36 @@ std::vector<Section> Section::sections(const util::Filter<Section>::type &filter
 }
 
 
+void addChildrenIfNotMaxDepth(std::tuple<Section, size_t>& current,
+                              std::list<std::tuple<Section, size_t>> &todo,
+                              size_t max_depth) {
+    if (std::get<1>(current) < max_depth) {
+        size_t next_depth = std::get<1>(current) + 1;
+        for (const auto &s : std::get<0>(current).sections()) {
+            todo.push_back(std::make_tuple(s, next_depth));
+        }
+    }
+}
+
+
 std::vector<Section> Section::findSections(const util::Filter<Section>::type &filter,
                                            size_t max_depth) const {
     std::vector<Section> results;
     std::list<std::tuple<Section, size_t>> todo;
-    bool first_iter = true;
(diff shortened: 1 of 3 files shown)


achilleas-k:
LOOKS AMAZING TO ME

src/Section.cpp
```diff
@@ -96,29 +96,26 @@ std::vector<Section> Section::sections(const util::Filter<Section>::type &filter
...
+            first_iter = false;
         }
-
         if (current.depth < max_depth) {
-            std::vector<Section> children = current.entity.sections();
             size_t next_depth = current.depth + 1;
-
-            for (auto it = children.begin(); it != children.end(); ++it) {
-                todo.push_back(SectionCont(*it, next_depth));
+            for (auto s : current.entity.sections()) {
```
gicmo: `const auto &s :`

src/Section.cpp
```diff
@@ -96,29 +96,26 @@ std::vector<Section> Section::sections(const util::Filter<Section>::type &filter
...
-        todo.pop_front();
-
-        bool filter_ok = filter(current.entity);
-        if (filter_ok) {
-            results.push_back(current.entity);
+    bool first_iter = true;
+
+    SectionCont current(*this);
+    while (todo.size() > 0 || first_iter) {
+        if(!first_iter) {
```
gicmo: I wonder if we need all that if stuff in the loop. Cant we just do 
```
for (const auto &s : this->sections()) {
 todo.push_back(SectionCont(s, next_depth));
}
```
before the loop? I think that might be a clearer.

achilleas-k: It would have to be
```c++
        if (current.depth < max_depth) {
            size_t next_depth = current.depth + 1;
            for (auto s : current.entity.sections()) {
                todo.push_back(SectionCont(s, next_depth));
            }
        }
```
but yes, we can. It was basically a choice between replicating the second half of the loop body outside or putting the first half behind a condition.

gicmo: Yeah, I personally think the not-branching-in-while-loop version is more readable.

gicmo:
While we are there, we could make `SectionCont` be like:
```
struct SectionCont {
    SectionCont(const Section &s, size_t d = 0)
    : entity(s), depth(d)
    {}

    const Section &entity;
    size_t depth;
};
```
I think that should work. Could also replace it with `std::tuple`, though that might make it less readable.

src/Section.cpp
```diff
@@ -76,56 +76,44 @@ bool Section::deleteSection(const Section &section) {
...
-        bool filter_ok = filter(current.entity);
-        if (filter_ok) {
-            results.push_back(current.entity);
+void addChildrenIfNotMaxDepth(std::tuple<Section, size_t>& current,
+                              std::list<std::tuple<Section, size_t>> &todo,
+                              size_t max_depth) {
+    if (std::get<1>(current) < max_depth) {
+        size_t next_depth = std::get<1>(current) + 1;
+        for (const auto &s : std::get<0>(current).sections()) {
+            todo.push_back(std::make_tuple(s, next_depth));
```
gicmo: `emplace_back`?

PR #694, opened 2017-09-03
Fix findRelated 

### LINKED ISSUE #691: 'Section.findRelated' 'findUpstream' vs 'findSideways'
The `nix.Section` method `findRelated` returns the first occurrence matching a filter criterion starting from any `Section` and returns it. To do this, it first searches downstream of the calling section, then upstream and then sideways.

The current implementation of `findUpstream` and `findSideways` are very similar. It looks like`findUpstream` already implements the "sideways" search. Maybe the `findUpstream` function could be refactored to speed up the upwards search or the "sideways" search was supposed to search sideways further down than it does currently.

jgrewe:
Due to our recent changes in findSection (exclude of the root section) the findRelated did not work properly in all cases (not caught by the tests :( ).
Extended the test cases and simplified findRelated, should fix #691

### DIFF
--- a/src/Section.cpp
+++ b/src/Section.cpp
@@ -127,6 +127,7 @@ static inline auto erase_section_with_id(std::vector<Section> &sections, const s
     return sections.size();
 }
 
+
 std::vector<Section> Section::findRelated(const util::Filter<Section>::type &filter) const
 {
     std::vector<Section> results = findDownstream(filter);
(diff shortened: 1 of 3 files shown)


src/Section.cpp
```diff
@@ -224,18 +222,18 @@ std::vector<Section> Section::findDownstream(const std::function<bool(Section)>
...
-        if (results.size() > 0) {
-            return results;
-        }
-        return p.findUpstream(filter);
+    if (p == none) {
+        return results;
+    }
+    if (filter(p)) {
+        results.push_back(p);
+        return results;
```
gicmo: So we always return exactly one? Or am I am missing something? You wanna keep the result type as a vector to stay compatible with the other search functions?

jgrewe: Yes, in this Case we return at maximum one section. And yes, I wanted to keep it similar to the other functions. Happy to change, if you want.

Issue #510, opened 2015-07-22
support compression of datasets


jgrewe:
- general filter API supporting custom filters/compressors
- overload the createDataArray methods to take the filter as optional argument
- if possible, implement the polynomial conversion as a filter working on the dataset

jgrewe:
compression available since #695

PR #695, opened 2017-09-06
Compression support

jgrewe:
addresses #510 

this is more a quick and dirty implementation of compression support. It uses a hdf5 built-in compression algorithm with a fixed compression level. At the moment, one can decide for each dataArray to compress it or not, default it true.

This pr is meant to get us thinking about how it should actually work. For the moment, I would prefer  such a simple solution which could be elaborated in future versions, or upon user request...

### DIFF
--- a/backend/fs/BlockFS.cpp
+++ b/backend/fs/BlockFS.cpp
@@ -23,16 +23,16 @@ BlockFS::BlockFS(const std::shared_ptr<base::IFile> &file, const std::string &lo
 
 
 BlockFS::BlockFS(const std::shared_ptr<base::IFile> &file, const std::string &loc, const std::string &id,
-                 const std::string &type, const std::string &name)
-    : EntityWithMetadataFS(file, loc, id, type, name)
+                 const std::string &type, const std::string &name, const Compression &compression)
+     : EntityWithMetadataFS(file, loc, id, type, name), compr(compression)
 {
     createSubFolders(file);
 }
 
 
 BlockFS::BlockFS(const std::shared_ptr<base::IFile> &file, const std::string &loc, const std::string &id,
-                 const std::string &type, const std::string &name, time_t time)
-    : EntityWithMetadataFS(file, loc, id, type, name, time)
+                 const std::string &type, const std::string &name, time_t time, const Compression &compression)
+     : EntityWithMetadataFS(file, loc, id, type, name, time), compr(compression)
 {
     createSubFolders(file);
 }
--- a/backend/fs/BlockFS.cpp
+++ b/backend/fs/BlockFS.cpp
@@ -250,7 +250,8 @@ bool BlockFS::deleteSource(const std::string &name_or_id) {
 //--------------------------------------------------
 
 std::shared_ptr<base::IDataArray> BlockFS::createDataArray(const std::string &name, const std::string &type,
-                                                           nix::DataType data_type, const NDSize &shape, bool compression) {
+                                                           nix::DataType data_type, const NDSize &shape,
+                                                           const Compression &compression) {
     if (name.empty()) {
         throw EmptyString("Block::createDataArray empty name provided!");
     }
--- a/backend/fs/BlockFS.cpp
+++ b/backend/fs/BlockFS.cpp
@@ -250,7 +250,7 @@ bool BlockFS::deleteSource(const std::string &name_or_id) {
 //--------------------------------------------------
 
 std::shared_ptr<base::IDataArray> BlockFS::createDataArray(const std::string &name, const std::string &type,
-                                                           nix::DataType data_type, const NDSize &shape) {
+                                                           nix::DataType data_type, const NDSize &shape, bool compression) {
     if (name.empty()) {
         throw EmptyString("Block::createDataArray empty name provided!");
     }
@@ -259,7 +259,7 @@ std::shared_ptr<base::IDataArray> BlockFS::createDataArray(const std::string &na
     }
     std::string id = util::createId();
     DataArrayFS da(file(), block(), data_array_dir.location(), id, type, name);
-    da.createData(data_type, shape);
+    da.createData(data_type, shape, compression);
     return std::make_shared<DataArrayFS>(da);
 }
 
--- a/backend/fs/DataArrayFS.cpp
+++ b/backend/fs/DataArrayFS.cpp
@@ -148,7 +148,7 @@ std::vector<double> DataArrayFS::polynomCoefficients() const {
 }
 
 
-void DataArrayFS::polynomCoefficients(const std::vector<double> &coefficients) {
+void DataArrayFS::polynomCoefficients(const std::vector<double> &coefficients, bool compression) {
     setAttr("polynom_coefficients", coefficients);
     forceUpdatedAt();
 }
@@ -239,8 +239,7 @@ bool DataArrayFS::deleteDimensions() {
 DataArrayFS::~DataArrayFS() {
 }
 
-
-void DataArrayFS::createData(DataType dtype, const NDSize &size) {
+void DataArrayFS::createData(DataType dtype, const NDSize &size, bool compression) {
     setDtype(dtype);
     dataExtent(size);
     /*
--- a/backend/fs/DataArrayFS.cpp
+++ b/backend/fs/DataArrayFS.cpp
@@ -148,7 +148,8 @@ std::vector<double> DataArrayFS::polynomCoefficients() const {
 }
 
 
-void DataArrayFS::polynomCoefficients(const std::vector<double> &coefficients, bool compression) {
+void DataArrayFS::polynomCoefficients(const std::vector<double> &coefficients,
+                                      const Compression &compression) {
     setAttr("polynom_coefficients", coefficients);
     forceUpdatedAt();
 }
@@ -239,7 +240,7 @@ bool DataArrayFS::deleteDimensions() {
 DataArrayFS::~DataArrayFS() {
 }
 
-void DataArrayFS::createData(DataType dtype, const NDSize &size, bool compression) {
+void DataArrayFS::createData(DataType dtype, const NDSize &size, const Compression &compression) {
     setDtype(dtype);
     dataExtent(size);
     /*
--- a/backend/hdf5/BlockHDF5.cpp
+++ b/backend/hdf5/BlockHDF5.cpp
@@ -290,7 +290,8 @@ shared_ptr<ITag> BlockHDF5::createTag(const std::string &name, const std::string
 shared_ptr<IDataArray> BlockHDF5::createDataArray(const std::string &name,
                                                   const std::string &type,
                                                   nix::DataType data_type,
-                                                  const NDSize &shape) {
+                                                  const NDSize &shape,
+                                                  bool compression) {
     string id = util::createId();
     boost::optional<H5Group> g = data_array_group(true);
 
@@ -298,7 +299,7 @@ shared_ptr<IDataArray> BlockHDF5::createDataArray(const std::string &name,
     auto da = make_shared<DataArrayHDF5>(file(), block(), group, id, type, name);
 
     // now create the actual H5::DataSet
-    da->createData(data_type, shape);
+    da->createData(data_type, shape, compression);
     return da;
 }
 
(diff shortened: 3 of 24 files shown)


achilleas-k:
Is there a reason we want compression defined for each DataSet separately? I imagine it would be nicer to have the option to enable during file creation and then each DataArray (or polynomial coefficient) creation method could check the global setting.

jgrewe:
@achilleas-k, well I first thought so as well but this would mean that we have to ask, e.g. ```File``` which mode was selected upon each creation of a ```DataSet```.... We can do it, though. I would keep the option to override the default and decide when not to compress, or vice versa. For example the imaging guys always keep the raw images uncompressed (well, image compression is something different, but there may be cases in which this is desired), or there are scenarios in which, e.g. for performance reasons, you want to have one ```DataArray``` uncompressed while you compress the others.

achilleas-k:
Yeah, I see. I can see how this might be useful.

Would it be too complicated if the user could specify the default value globally? In other words, when opening or creating a file, you could specify the true or false for compression, then the DataArray creation function accepts an argument that overrides that default.

Might be a bit messy though.

jgrewe:
I was wondering if such a default compression choice should be saved within the file? That is, when a file is re-opened do I automatically have the same settings? Such a feature would require a model change, however.

achilleas-k:
I guess it could be a top-level attribute (alongside file format and version), the absence of which signals "off". That way it would be backwards compatible in the file format.

jgrewe:
@achilleas-k tried to define a default during file opening. Unfortunately it won't work without an api change and this is something I would like to avoid.
We can set the default compression choice at the level of the File and pass it to the Block(s) which then can apply it at any place (in principle). The problem starts at the ````createDataArray``` function. If we want to avoid an api break, we have to add the compression flag with a default value. This is then leads to the situation that we always have a compression argument (with the default value) and a  maybe conflicting compression flag set at file opening. Maybe I am to blind, but I do not see a simple way to tell/decide which one  is the current user's wish. We could spend a further argument on this, but this becomes really ugly.

I would suggest to use the approach used here and default compression to a certain value that can be overridden at any place a dataset is created. For future versions that may break the the api anyway we may elaborate on this. Actually, I would really like to have the compression for our recordings, saves about 60% of disc space :)

achilleas-k:
I see. Yeah, I guess we can think about it again in the future. The global option would be in addition to the current change anyway, so we can go ahead with this one and work on it later (or not).

backend/fs/DataArrayFS.cpp
```diff
@@ -240,7 +240,7 @@ DataArrayFS::~DataArrayFS() {
 }
 
 
-void DataArrayFS::createData(DataType dtype, const NDSize &size) {
+    void DataArrayFS::createData(DataType dtype, const NDSize &size, bool compression) {
```
achilleas-k: Indent typo?

backend/hdf5/DataArrayHDF5.cpp
```diff
@@ -131,13 +131,13 @@ vector<double> DataArrayHDF5::polynomCoefficients() const {
...
 
-void DataArrayHDF5::polynomCoefficients(const vector<double> &coefficients) {
+void DataArrayHDF5::polynomCoefficients(const vector<double> &coefficients, bool compression) {
     DataSet ds;
     if (group().hasData("polynom_coefficients")) {
         ds = group().openData("polynom_coefficients");
         ds.setExtent({coefficients.size()});
     } else {
-        ds = group().createData("polynom_coefficients", H5T_NATIVE_DOUBLE, {coefficients.size()});
+         ds = group().createData("polynom_coefficients", H5T_NATIVE_DOUBLE, {coefficients.size()}, {}, {}, true, true, compression);
```
achilleas-k: Another one?

jgrewe: eh, fixed those ones, will rebase the branch once the other #696 is merged...

backend/hdf5/h5x/H5Group.cpp
```diff
@@ -217,8 +218,12 @@ DataSet H5Group::createData(const std::string &name,
         HErr res = H5Pset_chunk(dcpl.h5id(), rank, chunks.data());
         res.check("Could not set chunk size on data set creation plist");
     }
-
-    DataSet ds = H5Dcreate(hid, name.c_str(), fileType.h5id(), space.h5id(), H5P_DEFAULT, dcpl.h5id(), H5P_DEFAULT);
+    DataSet ds;
+    if (compression) {
+         HErr status = H5Pset_deflate (dcpl.h5id(), 6);
+         status.check("Could not set compression!");
```
gicmo: Which exception will this result in? Maybe we want to differentiate this use case client-side, so we a program could try to use compression, and if fails retry without it?

backend/hdf5/DataArrayHDF5.cpp
```diff
@@ -131,13 +131,13 @@ vector<double> DataArrayHDF5::polynomCoefficients() const {
...
 
-void DataArrayHDF5::polynomCoefficients(const vector<double> &coefficients) {
+void DataArrayHDF5::polynomCoefficients(const vector<double> &coefficients, bool compression) {
     DataSet ds;
     if (group().hasData("polynom_coefficients")) {
         ds = group().openData("polynom_coefficients");
         ds.setExtent({coefficients.size()});
     } else {
-        ds = group().createData("polynom_coefficients", H5T_NATIVE_DOUBLE, {coefficients.size()});
+        ds = group().createData("polynom_coefficients", H5T_NATIVE_DOUBLE, {coefficients.size()}, {}, {}, true, true, compression);
```
gicmo: We could (should?) re-order the arguments so we can make use of the default args most of the time. I am not sure about the rest of the codebase but I see a few new `{}, {}, true, true, compression` places. We could avoid that by moving the compression arg up behind size, but maybe then other places need more default args, not sure. Do you have an idea @jgrewe ? The `createData` is internal API so I think we can do that without too much hassle, if we wanted to.

backend/fs/BlockFS.cpp
```diff
@@ -250,7 +250,7 @@ bool BlockFS::deleteSource(const std::string &name_or_id) {
 //--------------------------------------------------
 
 std::shared_ptr<base::IDataArray> BlockFS::createDataArray(const std::string &name, const std::string &type,
-                                                           nix::DataType data_type, const NDSize &shape) {
+                                                           nix::DataType data_type, const NDSize &shape, bool compression) {
```
gicmo: Although I indeed very much like the simplicity of the approach, maybe we can be a tiny bit more extensible and future-proof by having an enum instead of the bool, ie:
```c++
enum class Compression {
   None,
   DeflateNormal,

   Auto = DeflateNormal,
   //and later:
   DeflateHigh,
   DeflateLow,

    FileDefault, // use the default store within the file
}
```

jgrewe:
@achilleas-k @gicmo, reworked the compression using @gicmo's enum approach. We can now pass it as either as a default while opening the file, or at each DataArray creation
Switched the default to NO compression

backend/hdf5/BlockHDF5.cpp
```diff
@@ -290,15 +292,16 @@ shared_ptr<ITag> BlockHDF5::createTag(const std::string &name, const std::string
...
+                                                  const Compression &compression) {
     string id = util::createId();
     boost::optional<H5Group> g = data_array_group(true);
 
     H5Group group = g->openGroup(name, true);
     auto da = make_shared<DataArrayHDF5>(file(), block(), group, id, type, name);
 
-    // now create the actual H5::DataSet
-    da->createData(data_type, shape);
+    // now create the actual H5::Data Set
```
gicmo: Nitpick: `H5::DataSet` was correct ;)

backend/hdf5/DataArrayHDF5.cpp
```diff
@@ -131,13 +131,13 @@ vector<double> DataArrayHDF5::polynomCoefficients() const {
...
 
-void DataArrayHDF5::polynomCoefficients(const vector<double> &coefficients) {
+void DataArrayHDF5::polynomCoefficients(const vector<double> &coefficients, const Compression &compression) {
     DataSet ds;
     if (group().hasData("polynom_coefficients")) {
         ds = group().openData("polynom_coefficients");
         ds.setExtent({coefficients.size()});
     } else {
-        ds = group().createData("polynom_coefficients", H5T_NATIVE_DOUBLE, {coefficients.size()});
+         ds = group().createData("polynom_coefficients", H5T_NATIVE_DOUBLE, {coefficients.size()}, compression);
```
gicmo: nitpick: did a space smuggle itself in?

backend/hdf5/h5x/H5Group.cpp
```diff
@@ -217,8 +218,12 @@ DataSet H5Group::createData(const std::string &name,
         HErr res = H5Pset_chunk(dcpl.h5id(), rank, chunks.data());
         res.check("Could not set chunk size on data set creation plist");
     }
-
-    DataSet ds = H5Dcreate(hid, name.c_str(), fileType.h5id(), space.h5id(), H5P_DEFAULT, dcpl.h5id(), H5P_DEFAULT);
+    DataSet ds;
+    if (compression == Compression::DeflateNormal) {
+         HErr status = H5Pset_deflate (dcpl.h5id(), 6);
+         status.check("Could not set compression!");
+    }
```
gicmo: Hmm, maybe it would be good to already add the case where an unsupported compression is specified (i.e. throw an exception)?

backend/hdf5/h5x/H5Group.hpp
```diff
@@ -47,14 +48,15 @@ class NIXAPI H5Group : public LocID {
...
+                       const NDSize &size,  const Compression &compression = Compression::Auto,
+                       const NDSize &maxsize = {}, NDSize chunks = {},
+                       bool maxSizeUnlimited = true, bool guessChunks = true) const;
 
     DataSet openData(const std::string &name) const;
     void removeData(const std::string &name);
 
     template<typename T>
-    void setData(const std::string &name, const T &value);
+         void setData(const std::string &name, const T &value, const Compression &compression = Compression::Auto);
```
gicmo: nitpick: new extra whitespaces?

include/nix/Compression.hpp
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2013, German Neuroinformatics Node (G-Node)
```
gicmo: nitpick: 2013 was quite some time ago ...

backend/hdf5/DataArrayHDF5.cpp
gicmo: Also more importantly: you really think it is necessary to compress the polynom dataset? Shouldn't those be only a few entries at all?

backend/hdf5/BlockHDF5.cpp
jgrewe: args, struggling with the emacs settings, in particular the indentation... Thought I found them all before pushing, apparently not.

backend/hdf5/DataArrayHDF5.cpp
jgrewe: actually not, that is because the header defines the default to be Compression::None

backend/hdf5/h5x/H5Group.cpp
jgrewe: I think that is exactly what ``check``does, As I see it, maybe I am wrong, ``HErr::check`` throws an [H5Error](https://github.com/G-Node/nix/blob/f352378c222c16d24e01f5c1f79398fb52fd1151/backend/hdf5/h5x/H5Object.hpp#L199)

backend/hdf5/BlockHDF5.cpp
gicmo: I think I fucked the emacs stuff up for us, fixating...

backend/hdf5/h5x/H5Group.cpp
jgrewe: or would you prefer an extra exception for just this case?

gicmo: What I meant is that, if we add a Compression::XXX in the future, we would just ignore it in that function here, right?

jgrewe:
a little rebase also from my side :)

PR #696, opened 2017-09-06
fix max_depth pass through and extend tests

### LINKED ISSUE #689: 'File.findSections' missing 'max_depth' passthrough
Currently the `nix.File` method `findSections(max_depth)` in `File.hpp` does not seem to pass `max_depth` to the executing function which leads to an unlimited search result.

jgrewe:
fixes #689


src/File.cpp
```diff
@@ -1,3 +1,4 @@
+
```
gicmo: Accident new newline?

jgrewe: ups

jgrewe: cf #695

PR #698, opened 2017-09-25
[build] optional static build of NIX

gicmo:
Use -DBUILD_STATIC on the cmake command line to enable a static
build of NIX, which includes linking boost and hdf5 statically
into a static version of NIX.


gicmo:
Did a quick rebase.

PR #704, opened 2017-09-27
Unit handling

jgrewe:
relaxed the unit handling at two places:

 1) ``Properties`` may have non-SI units. Passing an empty string, or one that consists of only blanks, will remove the unit from the property.
 2) ``DataArrays``: The value unit may be non-SI. In case the dimensionality of the ``DataArray`` is 1-D and the first dimension is an ``AliasRangeDimension`` setting an non-SI will lead to an exception.
In case a non-SI unit is set, the attempt of adding an ``AliasRangeDimension`` will lead to an exception.

Data retrieval functions behave as before: if there are units and theses can not be scaled, an exception will be thrown.

### DIFF
--- a/src/DataArray.cpp
+++ b/src/DataArray.cpp
@@ -8,7 +8,6 @@
 
 #include <nix/DataArray.hpp>
 
-#include <nix/util/util.hpp>
 #include "hdf5/h5x/H5DataType.hpp"
 
 #include <cstring>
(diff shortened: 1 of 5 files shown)


src/DataArray.cpp
```diff
@@ -99,9 +98,17 @@ void DataArray::appendData(DataType dtype, const void *data, const NDSize &count
 }
 
 void DataArray::unit(const std::string &unit) {
-    util::checkEmptyString(unit, "unit");
-    if (!unit.empty() && !(util::isSIUnit(unit) || util::isCompoundSIUnit(unit))) {
-        throw InvalidUnit("Unit is not SI or composite of SI units.", "DataArray::unit(const string &unit)");
+    std::string dblnk_unit = util::deblankString(unit);
+    util::checkEmptyString(dblnk_unit, "unit");
+    if (this->dimensionCount() == 1 && this->getDimension(1).dimensionType() == nix::DimensionType::Range) {
```
gicmo: That is quite a bit of nesting. Maybe we could invert it?

achilleas-k:
I'll allow it. 🙂

src/DataArray.cpp
jgrewe: better like this?
~~~c++

if (this->dimensionCount() != 1 || this.getDimension(1).dimensionType() != nix::DimensionType::Range) {
    backend()->unit(dblnk_unit);
    return;
}
nix::RangeDimension rd;
rd = this->getDimension(1);
if (rd.alias() && !(util::isSIUnit(dblnk_unit) || util::isCompoundSIUnit(dblnk_unit))) {
    throw InvalidUnit("Non-SI units are not allowed if the DataArray has an AliasRangeDimension. Data retrieval might not work!", "DataArray::unit(const string &unit)");
} else {
    backend()->unit(dblnk_unit);
    return;
}
~~~

gicmo: はい はい!

PR #707, opened 2017-10-19
Small h5x::DataType improvements

gicmo:
Split out from the DataFrame branch, to keep that PR as small as possible.

### DIFF
--- a/backend/hdf5/h5x/H5DataType.cpp
+++ b/backend/hdf5/h5x/H5DataType.cpp
@@ -16,6 +16,11 @@ namespace nix {
 namespace hdf5 {
 namespace h5x {
 
+bool DataType::equal(const DataType &other) const {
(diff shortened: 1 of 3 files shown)


backend/hdf5/h5x/H5DataType.cpp
```diff
@@ -16,6 +16,11 @@ namespace nix {
 namespace hdf5 {
 namespace h5x {
 
+bool DataType::equal(const DataType &other) const {
+    HTri res = H5Tequal(hid, other.hid);
+    res.check("DataType::equal(): H5Tequal failed");
+    return res.result();
+}
```
jgrewe: out of curiosity: why not the  ``==`` operator?

gicmo: Fair point. I guess I have been programming in plain C for too long. Added it.

PR #708, opened 2017-10-20
DataFrame [WIP]

gicmo:
This is a test balloon to see where if (I am sure it will) breaks on other platforms (win?).
It is mostly done. Mostly. Need more tests, as always.
Feedback on the API would also be good. Some bits are missing and some are to be discussed:
- Appending rows (currently done via `rows(NEW_SIZE)`)?
- FS backend and resizing of rows.

### DIFF
--- a/backend/hdf5/DataFrameHDF5.hpp
+++ b/backend/hdf5/DataFrameHDF5.hpp
@@ -54,6 +54,11 @@ public:
     Variant readCell(ndsize_t row, ndsize_t col);
     std::vector<Variant> readRow(ndsize_t row);
 
+    void writeColumn(ndsize_t col,
+                     ndsize_t offset,
+                     ndsize_t count,
+                     DataType dtype,
+                     const char *data);
 
 };
 
--- a/backend/hdf5/DataFrameHDF5.hpp
+++ b/backend/hdf5/DataFrameHDF5.hpp
@@ -58,13 +58,13 @@ public:
     void writeCells(ndsize_t row, const std::vector<Cell> &cells);
 
 
-    void readColumn(ndsize_t col,
+    void readColumn(int col,
                     ndsize_t offset,
                     ndsize_t count,
                     DataType dtype,
                     void *data) const;
 
-    void writeColumn(ndsize_t col,
+    void writeColumn(int col,
                      ndsize_t offset,
                      ndsize_t count,
                      DataType dtype,
--- a/backend/hdf5/DataFrameHDF5.hpp
+++ b/backend/hdf5/DataFrameHDF5.hpp
@@ -50,6 +50,8 @@ public:
     void writeCell(ndsize_t row, ndsize_t col, const Variant &v);
     void writeRow(ndsize_t row, const std::vector<Variant> &v);
 
+    Variant readCell(ndsize_t row, ndsize_t col);
+
 
 };
 
--- a/backend/hdf5/DataFrameHDF5.hpp
+++ b/backend/hdf5/DataFrameHDF5.hpp
@@ -45,8 +45,8 @@ public:
 
     std::vector<Column> columns() const override;
 
-    std::vector<unsigned> nameToCol(const std::vector<std::string> &names) const override;
-    std::vector<std::string> colToName(const std::vector<unsigned> &cols) const override;
+    std::vector<unsigned> colIndex(const std::vector<std::string> &names) const override;
+    std::vector<std::string> colName(const std::vector<unsigned> &cols) const override;
 
     ndsize_t rows() const override;
     void rows(ndsize_t n) override;
--- a/backend/hdf5/DataFrameHDF5.hpp
+++ b/backend/hdf5/DataFrameHDF5.hpp
@@ -60,6 +60,12 @@ public:
                      DataType dtype,
                      const char *data);
 
+    void readColumn(ndsize_t col,
+                    ndsize_t offset,
+                    ndsize_t count,
+                    DataType dtype,
+                    void *data);
+
 };
 
 
--- a/backend/hdf5/DataFrameHDF5.hpp
+++ b/backend/hdf5/DataFrameHDF5.hpp
@@ -12,6 +12,7 @@
 #define NIX_DATA_FRAME_HDF5_H
 
 
+#include <nix/NDSize.hpp>
 #include <nix/base/IDataFrame.hpp>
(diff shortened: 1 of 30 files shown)


backend/hdf5/DataFrameHDF5.hpp
```diff
@@ -0,0 +1,88 @@
...
+    DataFrameHDF5(const std::shared_ptr<base::IFile> &file, const std::shared_ptr<base::IBlock> &block, const H5Group &group, const std::string &id, const std::string &type, const std::string &name, time_t time);
+
+
+    void createData(const std::vector<Column> &cols, const Compression &compression);
+
+    std::vector<Column> columns() const override;
+
+    std::vector<unsigned> nameToCol(const std::vector<std::string> &names) const override;
+    std::vector<std::string> colToName(const std::vector<unsigned> &cols) const override;
+
```
jgrewe: I was initially confused by the names, at first sight thought ``nameToCol`` would actually create the column. 

what about names like ``colIndex(vector<string> colNames)`` and ``colNames(vector<unsigned> indices)``

include/nix/DataFrame.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+    Variant readCell(ndsize_t row, const std::string &col) const {
+        std::vector<Variant> cells = backend()->readCells(row, {col});
+        if (cells.size() < 1) {
+            return Variant{};
+        }
+
+        return cells[0];
+    }
+
```
jgrewe: would it make sense to offer a function to get the actual ``Cell`` (``vector<Cell> getCell(row, col)``)object?

backend/hdf5/DataFrameHDF5.hpp
gicmo: Sounds good.

include/nix/DataFrame.hpp
gicmo: I thought maybe `getCells(std::vector<Cell> &cells, ndsize_t row, std::vector<unsgined> cols, gboolean resize_vector = true)` ... but one that returns a new vector of `Cell`s could also work.

gicmo:
Todo List:
- [ ] Documentation
- [ ] Group's DataFrame access functions
- [ ] More tests
- [ ] API review and changes

include/nix/DataFrame.hpp
jgrewe: dunno, both versions seem fine to , except that returning of a new vector of Cells does not imply that the content is a reference to the actually stored cells...

include/nix/DataFrame.hpp
```diff
@@ -0,0 +1,142 @@
...
+        DataType dtype = hydra.element_data_type();
+        backend()->writeColumn(col, offset, count, dtype, (const char *) hydra.data());
+    }
+
+    template<typename T>
+        void readColumn(int col,
+                        ndsize_t offset,
+                        std::vector<T> &vals,
+                        ndsize_t count = 0) {
```
gicmo: API: This should include a `bool` allowing the vector to be resized to the requested `count` and a way to say "the whole column"

include/nix/DataFrame.hpp
```diff
@@ -0,0 +1,142 @@
...
+        Hydra<std::vector<T>> hydra(vals);
+
+        if (count == 0)
+            count = vals.size();
+
+        DataType dtype = hydra.element_data_type();
+        backend()->readColumn(col, offset, count, dtype, (void *) hydra.data());
+    }
+
```
gicmo: API: I guess we could expose the raw `readColumn`, `writeColumn` functions too? Or we can create a custom data structure that takes a raw pointer, length and datatype as a proxy structure. Might be nix also for DataArrays. Or both.

include/nix/DataFrame.hpp
jgrewe: what is the policy now? appending to the vector? maybe put the vector as second arguments and the others get defaults, count == 0 could indicate that the whole column should be read

jgrewe:
Do we also want to offer some sugar? Things like to and from csv? Can be added later on, of course.

include/nix/DataFrame.hpp
gicmo: currently we use the size of the vector, unless you specify count, then we assume the vector is big enough. We should at least make sure the latter is true I guess ;)

gicmo:
> Do we also want to offer some sugar? Things like to and from csv? Can be added later on, of course.

Yeah, I would add that later. But sounds not like a bad idea. The csv import/export could be used by the FS backend as well.

backend/hdf5/DataFrameHDF5.hpp
gicmo: Done.

include/nix/DataFrame.hpp
gicmo: I have added some commits with a new pair of functions. Maybe have a look if you like them.

gicmo:
Do we want a per-column "type" (not DataType, but as in the second parameter of the DataArray/Block/Group)? A "definition"?

achilleas-k:
Good question. I would expect the DataFrame to have a single type and description attribute like any other object, but since the data isn't homogeneous, it might useful to be able to specify a description for each. But it also gets a bit messy, doesn't it? Having attributes for each column in this way kinda breaks the idea of a unified table object.

I would lean towards not having type and description per column unless I hear a very compelling argument or example.

include/nix/DataFrame.hpp
gicmo: Maybe `Cell` should inherit from `Variant` and then all `getCell(s)` methods could just return `Cells`.

gicmo:
This is getting close. I guess some more tests are needed, but besides that it is basically ready. Review away ;)

gicmo:
> Is there any straightforward way to delete a specific row or column?

Column might be tricky, I am not sure we can alter it, from the docu:

> When a dataset is created, the storage datatype is defined. After the dataset or attri­bute is created, the datatype cannot be changed.

Rows of course should be doable. You think something like `DataFrame::deleteRow(ndsize_t index)`?

achilleas-k:
Yeah, delete row should be supported like you mentioned, just an index, or a range that deletes and shifts the rest "up".

As for column, yeah, I wasn't thinking implementation and deleting a common would be a pain. Probably easier to create a new one and copy the rest over, which is horrible.

PR #709, opened 2017-10-20
Version bump

jgrewe:


gicmo:
You want this to be merged to the master branch? Not the 1.4. branch? I think master should actually be bumped to say something like 1.5.0-pre or so.

jgrewe:
well, yes, this is what we did before. But you are right, strictly speaking the version number does not make sense for the development branch...

gicmo:
I would say bump it to 1.5.0-pre or something then (or 1.5.0.gitXXXX have to play around). See the [Fedora guidelines](https://fedoraproject.org/wiki/Packaging:Versioning) for a good start. Also we could create nixio.spec dynamically (with confgure_file from CMake), so so we dont have to do this by hand. And then have an extra copr for our dev snapshots.

jgrewe:
making the version something that is not convertible to int will break the apiVersion function, will bump it to 1.5.0, ok?

Issue #705, opened 2017-09-28
Print name when using nix-tool validate


mpsonntag:
For usability reasons it would be awesome to print not only the ID but also the name when `nix-tool validate` returns information.

gicmo:
That should be fixed the PR above ;)

mpsonntag:
`nix-tool validate` does not yet print the name of an entity in the windows build.

mpsonntag:
Closed for good via #713.

PR #710, opened 2017-10-22
Print entity name during validation

gicmo:
Fix issue #705 

My eyes still hurt from looking at some of the code ...


gicmo:
Oh, windows & vsc++ ... *crying* ...

Issue #595, opened 2016-02-04
Naming: keep NIX for the project, use "nixio" for the libraries


gicmo:
We currently have:
- NIX (C++)
  - homebrew science: `nixio`
- NixPy Python:
  -   PyPI: `nixpy` (?)

The idea would be to use `nixio` for all libraries to unify naming.

stoewer:
Agreed

jgrewe:
does that also mean that we change the names of the library?

libnix -> libnixio
import nixio
etc...

stoewer:
From my point of view this should definitely apply to:

* library names (so, dll, jar etc)
* package names (Debian, PyPi, homebrew)
* python package names (e.g. from nixio import File ...)

Optionally we can also change:

* C++ namespaces
* Main header (nix.hpp -> nixio.hpp)

jgrewe:
I am not sure about the namespaces but probably the header.
The project name itself will remain unchanged, correct?

stoewer:
Yes, the project and repository name(s) should remain unchanged

gicmo:
I think we should be done here, or can anybody think of anything else?

PR #719, opened 2018-01-14
[cmake] rename nix lib to nixio

gicmo:
Next step of the big rename (issue #595)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,21 +132,21 @@ else()
   set(NIX_LIBTYPE SHARED)
 endif()
 
-add_library(nix ${NIX_LIBTYPE} ${nix_INCLUDES} ${nix_SOURCES})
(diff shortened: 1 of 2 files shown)


jgrewe:
just wondered whether we also want to rename variables like ```nix_INCLUDES```  etc. ?

achilleas-k:
Is renaming the variables necessary? I suppose there's a consistency argument **for** renaming them. How far would those changes go? Build scripts, etc?

jgrewe:
I guess it is not necessary, my argument would be solely the consistency. The change should not affect anything else than the cmake file

gicmo:
Renaming things sounds fine to me, but I would suggest we wait with it until all the functional changes are done (i.e. the lib rename, and the nix tool rename). So we separate the aesthetics from the semantics.

PR #723, opened 2018-02-27
remove Values, move uncertainty to property fixes #550

### LINKED ISSUE #550: reduce unused odml values

jgrewe:
The property is now a dataset with a simple data type, no compound type anymore.

Implementation is based on the old value implementation. Maybe there are more elegant ways.
Value class still exists.
Feel free to speak your mind.


jgrewe:
travis fail of osx builds due to hdf5 install via home-brew. Has already been fixed in pr #722 , will leave it like this for the moment and rebase, once the #722 is merged

PR #727, opened 2018-04-29
Dimensions

jgrewe:
Add default arguments for the appendXYZDimension methods.

I do not remember if there was a decision against default values, but I find them quite convenient :) And they actually save me from writing very boring lines of code...

### DIFF
--- a/include/nix/DataArray.hpp
+++ b/include/nix/DataArray.hpp
@@ -293,10 +293,14 @@ public:
     /**
      * @brief Append a new SetDimension to the list of existing dimension descriptors.
      *
-     * @return The newly created SetDimension.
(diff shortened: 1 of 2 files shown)


include/nix/DataArray.hpp
```diff
@@ -349,11 +361,22 @@ class NIXAPI DataArray : public base::EntityWithSources<base::IDataArray>, publi
      * @brief Append a new SampledDimension to the list of existing dimension descriptors.
      *
      * @param sampling_interval         The sampling interval of the SetDimension to create.
+     * @param label                     The label of the dimension (e.g. 'time'), default: empty
+     * @param unit                      The unit of the dimension, default: empty
      *
```
achilleas-k: `@param offset` in doc?

PR #734, opened 2018-07-19
Proof of Concept of FileMode::Force

gicmo:
Gave it a quick shot, did not touch the FS backend. Might also have missed a spots.


achilleas-k:
@gicmo Should I merge this, or would you rather change the force option to a boolean?

gicmo:
I personally think we (I) should redo this to have a new `OpenFlags oflags` parameter instead.

PR #735, opened 2018-08-02
some modifications of doxygen settings, add generated api documentation to docs

jgrewe:
... and create a link in the index.md, let's see if it is really that simple


jgrewe:
Yes, agreed, have kind of bad feelings about this as well. On the contrary, this is basically what we did in the gh-pages branch. In that way we would have this as part of the GitHub.io pages, at least this is what I hope for

jgrewe:
Read a little bit about Jekyll. There are some hooks available but I did not find something that appeared appropriate for our purposes. 
Basically we would like to call a little script that clones the repo and calls make doc... or doxygen directly, which might be simpler. If anyone of you feels like doing some ruby... i have no clue.

PR #736, opened 2018-08-05
Implement data slice method

jgrewe:
... upon successful build and merge we may want to pick these changes to 1.4 as well.

### DIFF
--- a/src/util/dataAccess.cpp
+++ b/src/util/dataAccess.cpp
@@ -330,13 +330,13 @@ DataView dataSlice(const DataArray &array, const std::vector<double> &start, con
         throw UninitializedEntity();
     }
     if (start.size() == 0 || start.size() != end.size()) {
-        throw OutOfBounds("Number of start entries does not match number of end entries.", 0);
+        throw std::invalid_argument("Number of start entries does not match number of end entries.");
     }
     if (start.size() != array.dimensionCount()) {
-        throw OutOfBounds("Number of start/end entries does not match dimensionality of the data.", 0);
+        throw std::invalid_argument("Number of start/end entries does not match dimensionality of the data.");
     }
     if (units.size() > 0 && units.size() != start.size()) {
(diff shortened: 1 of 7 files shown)


jgrewe:
@gicmo and @achilleas-k: no more warnings, ready to merge?

achilleas-k:
I'm good to merge is @gicmo is happy with the changes he requested.

src/util/dataAccess.cpp
```diff
@@ -324,6 +324,43 @@ bool positionAndExtentInData(const DataArray &data, const NDSize &position, cons
...
+DataView dataSlice(const DataArray &array, const std::vector<double> &start, const std::vector<double> &end,
+                   const std::vector<std::string> &units) {
+    if (array == nix::none) {
+        throw UninitializedEntity();
+    }
+    if (start.size() == 0 || start.size() != end.size()) {
+        throw OutOfBounds("Number of start entries does not match number of end entries.", 0);
+    }
+    if (start.size() != array.dimensionCount()) {
+        throw OutOfBounds("Number of start/end entries does not match dimensionality of the data.", 0);
```
gicmo: We could pass `start.size()` as the "where" parameter (currently `0`). But maybe that is confusing.

gicmo: Actually, I think `OutOfBounds` is probably not 100% the right exception, but I don't think that is crucial.

gicmo:
Question is, do we bother finding better exceptions for the checks or not?

achilleas-k:
InvalidRank?

src/util/dataAccess.cpp
```diff
@@ -324,6 +324,43 @@ bool positionAndExtentInData(const DataArray &data, const NDSize &position, cons
...
+    }
+    if (units.size() > 0 && units.size() != start.size()) {
+        throw OutOfBounds("Number of units does not match dimensionality of the data.", 0);
```
gicmo: Technically also not `OutOfBounds`, I think.

PR #740, opened 2018-08-22
adds a little bit of sugar

jgrewe:
* **DataView** has a getter for the dimensions of the underlying data which simplifies working with data retrieved from a Tag/Mtag

* **MultiTag** has a positionCount method which simplifies reading the number of tagged positions

* **NameFilter** and **TypeFilter** can now be constructed with custom regular expressions. Not sure if it makes sense to add this to IdFilter too?

### DIFF
--- a/include/nix/MultiTag.hpp
+++ b/include/nix/MultiTag.hpp
@@ -112,6 +112,15 @@ public:
     }
 
     /**
+     * @brief Returns the number of tagged positions.
+     *
+     * @return The number of positions tagging the referenced data.
+     */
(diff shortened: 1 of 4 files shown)


achilleas-k:
What's our coveralls drop limit? 0.5%?

jgrewe:
I actually wonder how it could possibly drop that much . Even the line in databiew which is supposedly not tested is used in the testcard. If you insist I can try

achilleas-k:
Nah, I don't care. I was just curious if we have it set that low.

PR #743, opened 2018-08-27
Install headers in nixio-1.0 prefix

gicmo:
Should the last part of the renaming issue #595


gicmo:
NB: this might break the nix discovery in bindings that don't use pkg-config, i.e. Matlab (@mpsonntag) and Java.

PR #753, opened 2018-10-09
Change cmake to build shared and static libs in parallel

jgrewe:
This is the second step to solve the nix-mx build [see](https://github.com/G-Node/nix-mx/issues/173) issues. Idea is to have homebrew build and install static and shared libraries at the same time. So that the static nix lib can be found.

**DO NOT MERGE YET** 
(i) is  not yet done, (ii) some things are still weird

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ before_install:
     else
        export CTEST_ARGS=-VV;
     fi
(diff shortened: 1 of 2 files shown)


jgrewe:
closing because it is ugly and might not be necessary

PR #754, opened 2018-10-12
nix-tool only for shared library build

jgrewe:
this pr is part of the nix-mx, brew installation problem solution. This pr:
* disables the nix-tool build for the static build
* renames the static lib under win to allow for installing shared and static lib
* checks whether a pkgcfg file already exists, only if missing it is installed

We should be able to install static and shared lib in parallel without brew complaining about overwriting   links 🤞

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -247,13 +247,13 @@ endif()
 
 #pkg-config support
 configure_file(${CMAKE_SOURCE_DIR}/nix.pc.in ${CMAKE_BINARY_DIR}/nix.pc)
-install(FILES ${CMAKE_BINARY_DIR}/nix.pc DESTINATION ${LIB_INSTALL_DIR}/pkgconfig)
-
+if(NOT EXISTS "${LIB_INSTALL_DIR}/pkgconfig/nix.pc")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -237,24 +238,25 @@ endif()
...
+    RUNTIME DESTINATION bin)
+  install(DIRECTORY include/ DESTINATION ${INCLUDE_INSTALL_DIR})
+  install(DIRECTORY ${CMAKE_BINARY_DIR}/include/ DESTINATION ${INCLUDE_INSTALL_DIR})
+endif()
 
 #pkg-config support
 configure_file(${CMAKE_SOURCE_DIR}/nix.pc.in ${CMAKE_BINARY_DIR}/nix.pc)
-install(FILES ${CMAKE_BINARY_DIR}/nix.pc DESTINATION ${LIB_INSTALL_DIR}/pkgconfig)
-
+if(NOT EXISTS "${LIB_INSTALL_DIR}/pkgconfig/nix.pc")
```
gicmo: hmm, if we wanna package libnixio (debian, fedora) and ship the static and non-static version, then this might now work, because you build in a build-root that is separate from there rest of the system.
Maybe better would be a config option?

PR #755, opened 2018-10-29
[dataAccess] bind position and count dimension to the array extent ...

jgrewe:
this solves an issue in which a {1,1} dataset could not be read


jgrewe:
yes, this is true, but I agree the message was not good. Let me try again:

In cases that the data is 1D there is only a single dimension. When multi-tagging such data we require the tag positions array to be 2d (1st dim the number of points, second dim the number of dimension in the referenced data, 1d in this example). When reading the first position we must now read a slice form the positions array that is {1, 1}  which is not equal to {1}! (This was the case in the previous version and failed)

achilleas-k:
> yes, this is true, but I agree the message was not good. Let me try again:

Nah, the change was pretty clear from the message and the code change. Just wanted to make sure I was understanding the underlying issue correctly.

PR #757, opened 2018-12-24
data slice can live with less information

jgrewe:
with this pr, the ``util::dataSlice`` method automatically fills in missing information (to some extent). That is: start, end, and unit values may be missing or incomplete. If less info is given than there are dimensions defaults are assumed:
* ``units`` are filled with the ones defined in the dimension
* ``start`` is set to ``offset`` (``SampledDimension``), first tick (``RangeDimension``), or 0 (``SetDimension``)
* ``end`` is set to point (``SampledDimension``, ``RangeDimension``) or max index (``SetDimension``)

General rule: if information is provided, the first value given is assumed to belong to the first dimension the 2nd to the second, etc.

### DIFF
--- a/src/util/dataAccess.cpp
+++ b/src/util/dataAccess.cpp
@@ -257,9 +257,11 @@ void getOffsetAndCount(const MultiTag &tag, const DataArray &array, const vector
     vector<Dimension> dimensions = array.dimensions();
     vector<vector<double>> start_positions(dimensions.size());
     vector<vector<double>> end_positions(dimensions.size());
+
     for (size_t idx = 0; idx < indices.size(); ++idx) {
         temp_offset[0] = indices[idx];
         vector<double> offset, extent;
+
         positions.getData(offset, temp_count, temp_offset);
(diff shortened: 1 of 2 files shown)


src/util/dataAccess.cpp
```diff
@@ -325,32 +307,82 @@ bool positionAndExtentInData(const DataArray &data, const NDSize &position, cons
...
+            if (i >= units.size()) {
+                units.push_back(rd.unit() ? *rd.unit() : "none");
+            }
+        } else if (dt == DimensionType::Set) {
+            SetDimension sd = dim.asSetDimension();
+            if (i >= starts.size()) {
+                starts.push_back(0.0);
+            }
+            if (i >= ends.size()) {
+                ends.push_back(static_cast<double>(shape[i]-1));
```
gicmo: Maybe add a TODO/FIXME to make it clear we are aware that this might overflow the maximum representable integer (of `ndsize_t`) in `double`.

PR #758, opened 2018-12-29
overload RangeDimension::ticks to efficiently read subsets of ticks

jgrewe:
so far, we always read all ticks even if a subset was required.  added an overloaded version to avoid this. Chose the overload approach for downward comp and the easier binding. Open for other suggestions, however.

Basically the same problem exists for SetDimension. Currently we only support reading all labels.


backend/fs/DimensionFS.cpp
```diff
@@ -425,6 +425,22 @@ std::vector<double> RangeDimensionFS::ticks() const {
...
+    Group g = redirectGroup();
+    if (g.hasData("ticks")) {
+        g.getData("ticks", ticks);
+        return ticks;
+    } else if (g.hasData("data")) {
+        g.getData("data", ticks);
+        return ticks;
+    } else {
+        throw MissingAttr("ticks");
+    }
```
gicmo: Just kill the dead code?

backend/hdf5/DimensionHDF5.cpp
```diff
@@ -378,6 +378,22 @@ vector<double> RangeDimensionHDF5::ticks() const {
...
 
+vector<double> RangeDimensionHDF5::ticks(size_t start, size_t count) const {
+    vector<double> ticks;
+    ticks.resize(count);
+    H5Group g = redirectGroup();
+    std::string dset_name = g.hasData("ticks") ? "ticks" : "data";
+    DataSet ds = g.openData(dset_name);
+    h5x::DataType memType = data_type_to_h5_memtype(nix::DataType::Double);
+    DataSpace fileSpace, memSpace;
+    nix::NDSize o(1, start);
```
gicmo: Best variable name ever! (-;

src/Dimensions.cpp
```diff
@@ -377,14 +377,14 @@ void RangeDimension::ticks(const std::vector<double> &ticks) {
...
 double RangeDimension::tickAt(const ndsize_t index) const {
-
     size_t idx = check::fits_in_size_t(index, "Tick index exceeds memory (size larger than current system supports)");
 
-    vector<double> ticks = this->ticks();
-    if (idx >= ticks.size()) {
+    try {
+        vector<double> ticks = this->ticks(idx, 1);
+        return ticks[0];
+    } catch (...) {
```
gicmo: Ekks, wildcard exception catching... what if something that is "not out of bounds" goes wrong in any backend, this would convert that exception too.
I am always very sceptical of exceptions for control-flow (C++ is not python-ask-for-forgivness). For the `ticks[0]` access a simple `size()` check would suffice and the backend's `ticks(start, count)` function, since it is public, should handle out-of-bounds itself too (and throw the correct exception).

backend/hdf5/DimensionHDF5.cpp
```diff
@@ -378,6 +378,22 @@ vector<double> RangeDimensionHDF5::ticks() const {
...
+    nix::NDSize c(1, count);
+    std::tie(memSpace, fileSpace) = ds.offsetCount2DataSpaces(c, o);
+    ds.read(ticks.data(),  memType, memSpace, fileSpace);
```
gicmo: For out-of-bounds access, I think we would just get a generic `H5Error` exception (from [`DataSet::read`](https://github.com/G-Node/nix/blob/58389c5a9961ee0eb4a7272d0c158ad8ec74e1cb/backend/hdf5/h5x/H5DataSet.cpp#L31)). Maybe `H5Error::value` can tell us this is an out of bounds error and we can convert the exception to an `OutOfBounds` exception. Or, which I think I prefer, we do some manual bounds checking before accessing the data (maybe even in the front-end object) .

backend/fs/DimensionFS.cpp
```diff
@@ -425,6 +425,22 @@ std::vector<double> RangeDimensionFS::ticks() const {
...
+    */ // FIXME
+    return ticks;
```
gicmo: I guess we don't have tests for this yet :)

include/nix/Dimensions.hpp
```diff
@@ -624,14 +624,28 @@ class NIXAPI RangeDimension : public base::ImplContainer<base::IRangeDimension>
...
+     * The ticks map the index of the data at the respective dimension to other
+     * values. This can be used to store data that is sampled at irregular
+     * intervals.
+     *
+     * @param start size_t the start index.
+     * @param count size_t the number of ticks to read.
+     *
+     * @return A vector with the ticks for the dimension.
+     */
+    std::vector<double> ticks(size_t start, size_t count) const {
```
gicmo: A dimension can be of `ndsize_t`, right? So start could also be of `ndsize_t`. In theory `count` could to, but then it would not fit into `std::vector` (and maybe the RAM), so I think for `count` it is fine.

test/BaseTestDimension.cpp
```diff
@@ -401,6 +401,15 @@ void BaseTestDimension::testRangeTicks() {
...
 
+    ticks.resize(100);
+    for (size_t i = 0; i < 100; ++i) {
+        ticks[i] = i * 3.14;
+    }
+    rd.ticks(ticks);
+    CPPUNIT_ASSERT(rd.ticks(0, 10).size() == 10);
+    CPPUNIT_ASSERT(rd.ticks(1, 99).size() == 99);
+    CPPUNIT_ASSERT(rd.ticks(10, 1)[0] == 10 * 3.14);
+    CPPUNIT_ASSERT(rd.ticks(20, 1)[0] == 20 * 3.14);
```
gicmo: Maybe check some invalid access too?

include/nix/Dimensions.hpp
jgrewe: yes, that was my thinking, std::vector always uses size_t

backend/fs/DimensionFS.cpp
jgrewe: well, it is not actually dead, it would be needed, if we had that dataset stuff implemented in the FS backend. I can remove it if you insist

backend/fs/DimensionFS.cpp
jgrewe: correct :)

backend/hdf5/DimensionHDF5.cpp
jgrewe: you prefer offset?

gicmo: Nah I am ok with `o` and `c`.

backend/fs/DimensionFS.cpp
gicmo: I did not know that you were planning to have a `DataSet` abstraction in the FS backend. ... maybe our front-end/back-end separation is at the wrong level then... oh well. No no, I don't insist .. the FIXME could use some explanation though ;)

test/BaseTestDimension.cpp
jgrewe: some out of bounds test cases were added with  e1df84396992febb9e94254c63dd25ec7c7f1a1f

backend/hdf5/DimensionHDF5.cpp
jgrewe: doing some size checking before rushing into h5 exceptions. Throwing OutOfBounds now, when   size check fails

backend/fs/DimensionFS.cpp
jgrewe: fixmes got more text, not sure, when we will ever resolve them...

PR #760, opened 2019-01-04
[nixio.spec] add gcc-c++ to build requires, update changelog

jgrewe:


nixio.spec
```diff
@@ -57,6 +57,9 @@ make install DESTDIR=%{buildroot}
 
 
 %changelog
+* Thu Nov 01 2018 Jan Grewe <jan.grewe@g-node.org> - 1.4.0-1
+- version bump to 1.4.3
```
gicmo: The comment says `1.4.3`, the changelog header says `1.4.0-1` and the actual `Version:` says `1.5.0` ;)

jgrewe: meh, total mess

PR #766, opened 2019-02-09
deprecation of retrieve data functions ...

jgrewe:
replacement methods are now simply called ```featureData``` and ```taggedData```


achilleas-k:
Damn coveralls messing up our checks.

PR #771, opened 2019-04-14
WIP: Docker based CI

gicmo:


gicmo:
So, travis does not run as long as the PR is draft?

jgrewe:
don't know, apparently GitHub expects it to run, but on the travis side there is no trace of the pr...

PR #773, opened 2019-04-15
provides a better error message when opening fails

jgrewe:


backend/hdf5/FileHDF5.cpp
```diff
@@ -333,30 +333,45 @@ shared_ptr<base::IFile> FileHDF5::file() const {
...
         if (!root.getAttr("version", vv)) {
             check = false;
+            message = "File is not a valid NIX file, could not read version attribute!";
         } else {
             file_format_version = FormatVersion(vv);
             if (mode == FileMode::ReadWrite) {
                 check = my_version.canWrite(file_format_version);
+                if (!check) {
+                    message = "Cannot open file for ReadWrite access, format mismatch! ";
+                    std::stringstream stream;
```
gicmo: If you have a `stringstream` already, maybe better to use that for constructing the message instead of using `message.append`?

include/nix/Exception.hpp
```diff
@@ -90,7 +90,7 @@ class InvalidName: public std::invalid_argument {
 class InvalidFile: public std::invalid_argument {
 public:
     InvalidFile(const std::string &caller):
```
gicmo: It is not caller anymore, it is message, which actually makes more sense.

gicmo:
> Does anything speak against just throwing directly form `checkHeader`?

What I envisioned is `checkHeader (bool do_not_throw)` and then we call it with `checkHeader ((flags & OpenFlags::Force) != OpenFlags::Force)`. And in the function, if `do_not_throw` is false and we encounter an error we throw.

PR #780, opened 2019-08-26
github: schedule building of build-images

gicmo:
Every week build an image (based on ubuntu 18.04) which contains all the necessary dependencies to build nix.


achilleas-k:
What's the status of this PR?

gicmo:
We could still do that and then use the docker images in other projects. OTOH maybe it would be better to just generate nightly packages for the various platforms and use those. What do you think?

achilleas-k:
I think I would vote for nightly packages for the platforms.  If a docker image is needed (or more convenient) for some situation, then it can quickly be built from the distro package afterwards, so this way we get both.

gicmo:
Ok.

PR #784, opened 2019-09-04
Back to travis for CI

gicmo:
Github actions/workflows seem to not be ready for prime time, especially for cross PRs:
  - status seems to not being reported properly
  - secrets are (obviously) not being shared and therefore we would have to make the codecov.io token public
  - github actions need to be activated for the person/repo sending the PR


gicmo:
I wonder if we should disable valgrind tests because it takes forever and I am not sure anybody is actually looking at the results.

PR #791, opened 2019-10-09
[docs] NIX file validity and rules for validator

achilleas-k:
I was working on making sure the rules for the NIXPy validator match the one in this library and while reading through the validator code, writing down all the rules, I figured I might as well make available for documentation purposes.

I can imagine several target users for this document:
- Contributors/developers who don't want to violate any of the detailed rules of the NIX format.
- Users that might want be wondering why or how their NIX files are _invalid_ (according to the validator).
- Me! (it's mostly for me).

This was written based entirely on the validator code, so if there's anything that's wrong, it might also be a mistake in the validator itself.  There might also be some things that aren't expressed clearly, so let me know what you think.

Also not that I haven't linked to this file from anywhere else in the docs, so let me know if you want me to reference it from any of the other documents.


jgrewe:
one quick comment even without reading: Thank you!
The other thing is that you used markdown, but if we wanted to include it in the read-the-docs, we should use rst instead. I always wanted to remove the md files but kept them for reference ...

docs/validation.md
```diff
@@ -0,0 +1,93 @@
...
+
+## The NIX validator
+
+The NIXIO libraries implement a validator that can check a file for both soft and hard rule violations and inconsistencies.  The output of these validators label hard rule violations as _errors_ and soft rule violations as _warnings_.
+
+## Hard rules
+
+### File objects
+- Creation date must be set: HDF5 attribute `created_at`.
+
```
jgrewe: Don't we enforce format, and version attributes?

docs/validation.md
```diff
@@ -0,0 +1,93 @@
...
+### Block and Group objects
+- Rules for [Entities](#entities) apply.
+
+### DataArray objects
+- Rules for [Entities](#entities) apply.
+- Data type must be set: HDF5 attribute `dtype`.
+- Number of attached Dimension objects must match the number of dimensions in the data.
+- Number of Ticks in any attached RangeDimension objects must match the length of the corresponding dimension in the data.
+- Number of Labels in any attached SetDimension objects must match the length of the corresponding dimension in the data.
```
jgrewe: if given, Labels may be empty in a Set Dimension

docs/validation.md
```diff
@@ -0,0 +1,93 @@
...
+- Position must be set: HDF5 dataset `position`.
+- If the Tag contains references, the length of `position` must match the number of dimensions in all the referenced DataArray objects.
+- If the Tag contains an extent, the length of `extent` must match the number of dimensions in all the referenced DataArray objects (and the length of `position`).
+- Units must be set to an atomic SI unit: HDF5 dataset `unit`.
+- Number of units must match the number of dimensions in all the referenced DataArray objects.
+- Units must be convertible to the corresponding units in all the referenced DataArray objects.
+
+### MultiTag objects
+- Rules for [Tag objects](#tag-objects) apply.
+- Number of Positions must match number of Extents.
```
jgrewe: maybe invert the sentence because extents are optional

docs/validation.md
achilleas-k: Right.  I think I skipped them because the file wont even open without them, but they should be mentioned.

achilleas-k: Oh no, I just realised.  Format and Version appear in the Soft Rules below.

Here they're marked as `should`:
https://github.com/G-Node/nix/blob/789e1bc802fd22470c0c5e54c789400408ee5d4b/src/valid/validate.cpp#L233-L234

gicmo:
Nice!

PR #799, opened 2019-11-01
Column dim implementation

jgrewe:
Not entirely done yet, e.g. tests are missing. But please feel free to comment.

### DIFF
--- a/backend/hdf5/DataArrayHDF5.cpp
+++ b/backend/hdf5/DataArrayHDF5.cpp
@@ -206,9 +206,9 @@ std::shared_ptr<base::ISampledDimension> DataArrayHDF5::createSampledDimension(n
 }
 
 
-std::shared_ptr<base::IColumnDimension> DataArrayHDF5::createColumnDimension(ndsize_t index, const nix::DataFrame &df, unsigned col_index) {
+std::shared_ptr<base::IColumnDimension> DataArrayHDF5::createColumnDimension(ndsize_t index, const nix::DataFrame &df, std::vector<unsigned> col_indices) {
     H5Group g = createDimensionGroup(index);
-    return make_shared<ColumnDimensionHDF5>(g, index, file(), block(), df, col_index);
+    return make_shared<ColumnDimensionHDF5>(g, index, file(), block(), df, col_indices);
 }
 
 H5Group DataArrayHDF5::createDimensionGroup(ndsize_t index) {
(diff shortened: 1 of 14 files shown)


jgrewe:
I do have some questions about the API:

1. do we really want to support using multiple columns? Imho, yes. Do we support complete rows? How?
2. what convenience functions should be offered on the front-end side? E.g. getters allowing for using of col names and indices.
3. Are the indices in getters the indices in the referenced DataFrame, or is front-end indexing relative to the number of indices used for the columnDimension?

things that need to be done: 

- implement tests
- update docstrings  (some have not been updated to reflect the option of multiple columns)
- documentation and examples for read the docs
- Wiki update regarding columnDimension and DataFrame

backend/hdf5/DimensionHDF5.cpp
```diff
@@ -259,13 +268,118 @@ void SetDimensionHDF5::labels(const none_t t) {
...
+std::vector<std::string> ColumnDimensionHDF5::labels() const {
+    std::vector<unsigned> col_indices = columnIndices();
+    nix::DataFrame df = dataFrame();
+    return df.colName(col_indices);
+}
+
+Column ColumnDimensionHDF5::column(unsigned col_index) const {
+    std::vector<nix::Column> cols = this->dataFrame()->columns();
+    return cols[col_index];
+}
```
achilleas-k: This returns the Column from the referenced DataFrame at a given Index.  Is it not affected by the `columnIndices` of the ColumnDimension object in any way.  Is it a convenience function for retrieving arbitrary columns from the DataFrame?

backend/hdf5/DataArrayHDF5.cpp
```diff
@@ -206,6 +206,11 @@ std::shared_ptr<base::ISampledDimension> DataArrayHDF5::createSampledDimension(n
 }
 
 
+std::shared_ptr<base::IColumnDimension> DataArrayHDF5::createColumnDimension(ndsize_t index, const nix::DataFrame &df, std::vector<unsigned> col_indices) {
+    H5Group g = createDimensionGroup(index);
+    return make_shared<ColumnDimensionHDF5>(g, index, file(), block(), df, col_indices);
+}
```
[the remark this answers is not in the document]
jgrewe: yes, in the frontend we deprecated them but kept them in the backend

backend/hdf5/DimensionHDF5.cpp
jgrewe: good point, so far this function would return a column even if this is not used in the column dimension. This issue is related to my question No 3. But actually your comment indicates that the index here should be an index in the list of used indices

achilleas-k: Yeah I just noticed you asked similar questions in your comment and I admit I forgot you had a comment with questions in the PR.  Sorry about that.

jgrewe:
closing since it will be replaced with the DataFrameDimension as in nippy

PR #801, opened 2019-11-26
Tag flexibility

jgrewe:
addresses #800 but only for 1.4

PR #805, opened 2020-03-17
Tag flexibility

jgrewe:
companion to #801 but for master branch

### DIFF
--- a/src/util/dataAccess.cpp
+++ b/src/util/dataAccess.cpp
@@ -245,14 +245,12 @@ void getOffsetAndCount(const Tag &tag, const DataArray &array, NDSize &offset, N
 
     if (units.size() == 0) {
         units = std::vector<std::string>(position.size(), "none");
-    } else if (units.size() > position.size()) {
-        while (units.size() > position.size()) {
-            units.pop_back();
-        }
-    } else {
-        while (units.size() < position.size()) {
-            units.push_back(getDimensionUnit(dimensions[units.size()]));
-        }
+    }
+    while (units.size() > position.size()) {
+        units.pop_back();
+    }
+    while (units.size() < position.size()) {
+        units.push_back(getDimensionUnit(dimensions[units.size()]));
     }
 
     NDSize temp_offset(position.size());
--- a/src/util/dataAccess.cpp
+++ b/src/util/dataAccess.cpp
@@ -269,29 +269,24 @@ void getOffsetAndCount(const MultiTag &tag, const DataArray &array, const vector
     DataArray extents = tag.extents();
     NDSize position_size, extent_size;
     ndsize_t dimension_count = array.dimensionCount();
+    vector<Dimension> dimensions = array.dimensions();
     vector<string> units = tag.units();
 
     while (units.size() < dimension_count) {
         units.push_back("none");
+    } // ok
+
+    vector<pair<double, double>> max_extents(dimensions.size());
+    for (size_t i = 0; i < dimensions.size(); ++i) {
+        double pos, ext;
+        getMaxExtent(dimensions[i], array.dataExtent()[i]-1, pos, ext);
+        max_extents.emplace_back(pos, ext);
     }
     if (positions) {
         position_size = positions.dataExtent();
     }
     if (extents) {
         extent_size = extents.dataExtent();
-    }
-
-    if (position_size.size() == 1 && dimension_count != 1) {
-        throw IncompatibleDimensions("Number of dimensions in positions does not match dimensionality of data",
-                                          "util::getOffsetAndCount");
-    }
-    if (position_size.size() > 1 && position_size[1] > dimension_count) {
-        throw IncompatibleDimensions("Number of dimensions in positions does not match dimensionality of data",
-                                          "util::getOffsetAndCount");
-    }
-    if (extents && extent_size != position_size) {
-        throw IncompatibleDimensions("Number of dimensions in extents does not match dimensionality of data",
-                                          "util::getOffsetAndCount");
     }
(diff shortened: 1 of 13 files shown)


src/valid/checks.cpp
```diff
@@ -88,14 +88,15 @@ bool extentsMatchRefs::operator()(const DataArray &extents) const {
...
+    /*
     auto extSize = extents.size();
     auto it = refs.begin();
     while (!mismatch && (it != refs.end())) {
         auto arrayExtent = (*it).dataExtent();
         mismatch = extSize != arrayExtent.size();
         ++it;
     }
-
+    */
```
achilleas-k: Saw this in the original PR too.  Do you want to keep the old code around for some reason?

jgrewe: yes, thought I want to keep it for reference. Thought one might want to have a validity check of some sort...

gicmo:
Now there are merge conflicts and the new `DataFrameDimension` is missing everywhere ;P

src/valid/checks.cpp
gicmo: I am not against keeping it in a comment, but then we should add a comment why it is there.

src/util/dataAccess.cpp
```diff
@@ -165,21 +165,114 @@ ndsize_t positionToIndex(double position, const string &unit, const RangeDimensi
...
+        break;
+    }
+    case DimensionType::Sample: {
+        SampledDimension sd = dim.asSampledDimension();
+        unit = sd.unit() ? *sd.unit() : "none";
+        break;
+    }
+    case DimensionType::Range: {
+        RangeDimension rd = dim.asRangeDimension();
+        unit = rd.unit() ? *rd.unit() : "none";
```
gicmo, line 226: `rd.unit.value_or("none")`

src/util/dataAccess.cpp
```diff
@@ -165,21 +165,114 @@ ndsize_t positionToIndex(double position, const string &unit, const RangeDimensi
...
+    if (extent.size() == 0) {
+        extent.resize(position.size(), 0.0);
+    }
+    while (position.size() > dim_count) {
+        position.pop_back();
+        extent.pop_back();
+    }
+    while (position.size() < dim_count) {
+        position.push_back(get<0>(max_extents[position.size()]));
+        extent.push_back(get<1>(max_extents[extent.size()]));
```
gicmo, line 261: I am curious if the compiler is smart enough to optimize the access of the pair.

src/util/dataAccess.cpp
jgrewe: Citing boost: ``Effects: Equivalent to if (*this) return **this; else return std::forward<U>(v);``

but easier to read, true

gicmo: I was not implying your code was wrong ;)

src/util/dataAccess.cpp
jgrewe: you got a suggestion?

gicmo: Nothing obvious, I think.

src/util/dataAccess.cpp
jgrewe: got that, and I agree that you suggestion is easier to read but since I came to like the ternary operator...
When you got a hammer, everything looks like a nail ...

jgrewe:
@achilleas-k @gicmo, I guess I am done with it for now (if the checks pass)

test/TestDataType.cpp
```diff
@@ -33,16 +33,16 @@ void TestDataType::setUp() {
...
-                   {nix::DataType::Int16, 2},
-                   {nix::DataType::UInt16, 2},
-                   {nix::DataType::Int32, 4},
-                   {nix::DataType::UInt32, 4},
-                   {nix::DataType::Float, 4},
-                   {nix::DataType::Int64, 8},
-                   {nix::DataType::UInt64, 8},
-                   {nix::DataType::Double, 8}};
+    fixed_sizes = decltype(fixed_sizes) {{nix::DataType::Int8, 1},
+                                         {nix::DataType::Int8, 1},
```
achilleas-k, line 37: Were you just changing the indentation here and accidentally dropped the `U`, or was this intentional?

jgrewe: good catch!

achilleas-k:
Is this good to go?

gicmo:
> Is this good to go?

I managed to review the code, and that is fine, I have not yet a chance to think through what it actually does. I try again tonight or will default to 'yes' ;)

src/util/dataAccess.cpp
```diff
@@ -165,21 +165,114 @@ ndsize_t positionToIndex(double position, const string &unit, const RangeDimensi
...
+        ext = sd.positionAt(max_index);
+    } else if (dt == DimensionType::Range) {
+        RangeDimension rd = dim.asRangeDimension();
+        pos = rd.tickAt(0);
+        ext = rd.tickAt(max_index);
+    } else if (dt == DimensionType::Set) {
+        SetDimension sd = dim.asSetDimension();
+        pos = 0.0;
+        if (max_index > pow(FLT_RADIX, std::numeric_limits<double>::digits)) {
+            throw nix::OutOfBounds("dataAccess::fillPositionsExtents: shape cannot be cast to double without loss of precision. Please open an issue on github!");
```
gicmo, line 182: This took me a bit, but maybe for similarly incapable floating point and math "dummies" as I am, a comment would be nice. 
Also, I think we can indeed make this a compile time constant by doing `(1 << std::numeric_limits<double>::digits)`

Issue #775, opened 2019-05-07
Add UUIDs to File objects


achilleas-k:
Last item left to finish with Issue #657 is to add UUIDs to files, since odML has document-level UUIDs. At first this will be an easy backwards-compatible change, but if we start relying on file UUIDs for any functionality (maybe external tooling?) it will become necessary.

jgrewe:
closed via #806

PR #806, opened 2020-03-24
File ID solves #775

jgrewe:


backend/hdf5/FileHDF5.cpp
```diff
@@ -371,6 +385,14 @@ bool FileHDF5::checkHeader(FileMode mode, bool throw_error) {
         check = false;
         message << "File is not a valid NIX file, version attribute missing!";
     }
+
+    if (check && my_version >= file_format_version) {
```
achilleas-k:
```suggestion
    if (check && my_version >= nix::FormatVersion({1, 2, 0})) {
```

We should check against `1.2.0`, no?  The version that introduced the ID.

jgrewe: good catch, but actually it should be 
```file_format_version >= nix::FormatVersion({1, 2, 0})```
correct?

PR #807, opened 2020-03-27
DataFrameDimension

jgrewe:


achilleas-k:
LGTM.  Got a couple of issues for NIXPy out of this.

achilleas-k:
@gicmo Will you have a look at this or should I go ahead and merge?
Not sure what's up with codecov and if we should try to restart it somehow until it's finished.

backend/hdf5/DimensionHDF5.cpp
```diff
@@ -259,13 +269,153 @@ void SetDimensionHDF5::labels(const none_t t) {
...
 
+//--------------------------------------------------------------
+// Implementation of DataFrameDimensionHDF5
+//--------------------------------------------------------------
+DataFrameDimensionHDF5::DataFrameDimensionHDF5(const H5Group &group, ndsize_t index)
+    : DimensionHDF5(group, index)
+{
+    setType();
+    // Note: maybe having this is a bad idea, because a DataFrameDimension created
+    // this way will not have the block and file
```
gicmo, line 280: Do we need it, if not remove it?

backend/hdf5/DimensionHDF5.cpp
```diff
@@ -259,13 +269,153 @@ void SetDimensionHDF5::labels(const none_t t) {
...
+}
+
+DataFrameDimensionHDF5::DataFrameDimensionHDF5(const H5Group &group, ndsize_t index,
+                                               const std::shared_ptr<IFile> &file,
+                                               const std::shared_ptr<IBlock> &block, const DataFrame &frame)
+    :DataFrameDimensionHDF5(group, index, file, block)
+{
+    std::shared_ptr<IDataFrame> idf = block->getEntity<IDataFrame>(frame.id());
+    if (!idf)
+        throw std::runtime_error("ColumnDimensionHDF5 DataFrame not found in block!");
```
gicmo, line 298: `ColumnDimensionHDF5` -> `DataFrameDimensionHDF5`

backend/hdf5/DimensionHDF5.cpp
```diff
@@ -259,13 +269,153 @@ void SetDimensionHDF5::labels(const none_t t) {
...
+}
+
+int DataFrameDimensionHDF5::checkColumnIndex(int col_index) const {
+    int column_index = col_index;
+    if (col_index == -1) {
+        column_index = columnIndex();
+    }
+
+    if (column_index == -1) {
+        throw nix::OutOfBounds("DataFrameDimension: Error accessing column, no column index was given and no default is specified.");
```
gicmo, line 331: Sometimes the error have `HDF5` suffix, sometimes not.

backend/hdf5/DimensionHDF5.cpp
jgrewe: maybe help me answering the question. RangeDimension and also the DataFrameDimension rely on file and block, but all Dimensions have this kind of constructor. 
I would rather remove the constructors.

backend/hdf5/DimensionHDF5.cpp
jgrewe: meh, good catch

backend/hdf5/DimensionHDF5.cpp
jgrewe: done, all get suffixes

backend/hdf5/DimensionHDF5.cpp
jgrewe: we can not remove the constructors, removed the note

backend/hdf5/DimensionHDF5.cpp
```diff
@@ -259,13 +269,153 @@ void SetDimensionHDF5::labels(const none_t t) {
...
+
+
+std::string DataFrameDimensionHDF5::label(int col_index) const {
+    int column_index;
+    nix::DataFrame df = dataFrame();
+    try {
+        column_index = checkColumnIndex(col_index);
+    } catch (const nix::OutOfBounds& e) {
+        if (col_index == -1)
+            return df.name();
```
gicmo, line 365: I think, if we just move that above `column_index = checkColumnIndex(col_index);` we can get rid of the `try-catch` block. I am not a huge fan of using exceptions for control flow in C++.

backend/hdf5/DimensionHDF5.cpp
```diff
@@ -259,13 +269,153 @@ void SetDimensionHDF5::labels(const none_t t) {
...
+std::string DataFrameDimensionHDF5::unit(int col_index) const {
+    int column_index = checkColumnIndex(col_index);
+
+    nix::DataFrame df = dataFrame();
+    std::vector<Column> cols = df.columns();
+    return cols[column_index].unit;
+}
+
+
+std::string DataFrameDimensionHDF5::label(int col_index) const {
```
gicmo, line 358: if `int` is used as a way to indicate that you might not actually want to select a column, maybe `boost::optional<unsigned>` would be better, so we don't have to mix types between signed and unsigned.

include/nix/DataArray.hpp
```diff
@@ -381,6 +382,84 @@ class NIXAPI DataArray : public base::EntityWithSources<base::IDataArray>, publi
...
+     */
+    DataFrameDimension appendDataFrameDimension(const DataFrame &frame, const std::string &column_name) {
+        if (!frame) {
+            throw nix::UninitializedEntity();
+        }
+        unsigned index;
+        try {
+            std::vector<std::string> names = {column_name};
+            std::vector<unsigned> indices = frame.colIndex(names);
+            index = indices[0];
```
gicmo, line 433: There is `unsigned colIndex(const std::string &name) const` as a nonThere is `unsigned colIndex(const std::string &name) const` as a non-vector overload.

backend/hdf5/DimensionHDF5.cpp
```diff
@@ -317,7 +470,7 @@ boost::optional<std::string> RangeDimensionHDF5::label() const {
 void RangeDimensionHDF5::label(const string &label) {
     H5Group g = redirectGroup();
     g.setAttr("label", label);
-    // NOTE: forceUpdatedAt() not possible since not reachable from here
+    // NOTE: forceUpdatedAt() not possible since not reacha?ble from here
```
gicmo, line 473: `reacha?ble`

.travis.yml
```diff
@@ -28,6 +28,10 @@ matrix:
       env: FS_BACKEND=OFF DOCKERFILE=ub1804
 
 addons:
+  apt:
+    update: true
+    packages:
+      - lcov
```
gicmo, line 34: That is in theory not needed anymore, right @achilleas-k

backend/hdf5/DimensionHDF5.cpp
```diff
@@ -259,13 +269,156 @@ void SetDimensionHDF5::labels(const none_t t) {
...
+    if (!col_index) {
+        column_index = columnIndex();
+    }
+
+    if (!column_index) {
+        throw nix::OutOfBounds("DataFrameDimensionHDF5: Error accessing column, no column index was given and no default is specified.");
+    }
+
+    nix::DataFrame df = dataFrame();
+    std::vector<Column> cols = df.columns();
```
gicmo, line 337: Krass oversight in the `DataFrame` API that we have to fetch all the column information if all we want is the number. Filed as issue #808

backend/hdf5/DimensionHDF5.cpp
```diff
@@ -259,13 +269,156 @@ void SetDimensionHDF5::labels(const none_t t) {
...
+    return cols[*column_index];
+}
+
+
+nix::DataType DataFrameDimensionHDF5::columnDataType(boost::optional<unsigned> col_index) const {
+    boost::optional<unsigned> column_index = checkColumnIndex(col_index);
+
+    nix::DataFrame df = dataFrame();
+    std::vector<Column> all_cols = df.columns();
```
gicmo, line 391: Also lame that we have to fetch all of them if we only want one. Filed as issue #809

.travis.yml
achilleas-k: Correct.  Remove it.

gicmo: No U!

PR #810, opened 2020-04-05
UTF-8 encoding of all strings for names and contents

### LINKED ISSUE #779: UTF-8 for strings
By default strings are encoded as [ASCII](https://support.hdfgroup.org/HDF5/doc/Advanced/UsingUnicode/index.html)  in HDF5. NIXPY already marks strings as UTF-8 (because h5py automatically does so) and for 1.5 I think we should do the same in NIX. From the aforementioned doc:

```
lcpl_id =    H5Pcreate(H5P_LINK_CREATE) ;
error =      H5Pset_char_encoding(lcpl_id, H5T_CSET_UTF8) ;
dataset_id = H5Dcreate2(group_id, "datos_ñ", datatype_id, dataspace_id, lcpl_id, H5P_DEFAULT, H5P_DEFAULT) ;               
```

gicmo:
This enables utf8 encoding for the names of groups, datasets and links as well as for all string values in datasets and attributes.
Closes #779

### DIFF
--- a/backend/hdf5/h5x/H5DataType.cpp
+++ b/backend/hdf5/h5x/H5DataType.cpp
@@ -34,10 +34,11 @@ DataType DataType::make(H5T_class_t klass, size_t size) {
     return dt;
 }
 
-DataType DataType::makeStrType(size_t size) {
+DataType DataType::makeStrType(size_t size, H5T_cset_t cset) {
     DataType str_type = H5Tcopy(H5T_C_S1);
     str_type.check("Could not create string type");
     str_type.size(size);
+    str_type.cset(cset);
     return str_type;
 }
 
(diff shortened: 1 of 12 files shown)


backend/hdf5/h5x/H5DataType.cpp
```diff
@@ -76,6 +77,15 @@ H5T_sign_t DataType::sign() const {
     return res;
 }
 
+void DataType:: cset(H5T_cset_t cset) {
```
jgrewe, line 80: stray blank here? `DataType:: cset()`
no blank lines between setter and getter?

gicmo:
Almost good to go, I want another tests that checks all the link names, attributes and contents have the correct character set.

jgrewe:
lgtm, @gicmo will you rebase it or should I doe the update here?

PR #813, opened 2020-04-15
Change creation of Properties

jgrewe:
So far we set chunking to auto and hence properties reserved quite a bit of unnecessary overhead. With this PR, the default size is reduced to 8 entries for Properties that are created without or only a single value. If they are created with a vector of values, the vector size is used to define the size and the chunking of the H5 dataset. 
Compression has been switched on.

These changes are analogous to recent nixpy changes

### DIFF
--- a/backend/hdf5/SectionHDF5.cpp
+++ b/backend/hdf5/SectionHDF5.cpp
@@ -254,7 +254,8 @@ shared_ptr<IProperty> SectionHDF5::getProperty(ndsize_t index) const {
 shared_ptr<IProperty> SectionHDF5::createProperty(const string &name, const DataType &dtype, const NDSize &shape) {
     string new_id = util::createId();
     boost::optional<H5Group> g = property_group(true);
-    DataSet ds = g->createData(name, data_type_to_h5_filetype(dtype), shape);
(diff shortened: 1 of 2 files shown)


backend/hdf5/SectionHDF5.cpp
```diff
@@ -251,24 +251,32 @@ shared_ptr<IProperty> SectionHDF5::getProperty(ndsize_t index) const {
...
     boost::optional<H5Group> g = property_group(true);
-    DataSet ds = g->createData(name, data_type_to_h5_filetype(dtype), {0});
+    DataSet ds = g->createData(name, data_type_to_h5_filetype(dtype), shape, Compression::DeflateNormal,
+                               {}, shape, true, false);
     return make_shared<PropertyHDF5>(file(), ds, new_id, name);
 }
 
 
+shared_ptr<IProperty> SectionHDF5::createProperty(const string &name, const DataType &dtype) {
+    shared_ptr<IProperty> p = createProperty(name, dtype, {8});
```
gicmo, line 264: why `8`? I would suggest a define, `DEFAULT_PROPERTY_SIZE`, with some comment about the trade offs,

jgrewe: haha, fair enough, will do

PR #814, opened 2020-04-16
Data access df dim

jgrewe:
not completely done yet. Needs a bit more testing and  support for column if a default unit is specified

### DIFF
--- a/src/Dimensions.cpp
+++ b/src/Dimensions.cpp
@@ -255,7 +255,11 @@ vector<double> SampledDimension::axis(const ndsize_t count, const ndsize_t start
     double offset =  backend()->offset() ? *(backend()->offset()) : 0.0;
     double sampling_interval = backend()->samplingInterval();
     for (size_t i = 0; i < axis.size(); ++i) {
-        axis[i] = (static_cast<double>(i) + startIndex) * sampling_interval + offset;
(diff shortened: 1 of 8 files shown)


src/Dimensions.cpp
```diff
@@ -255,7 +255,11 @@ vector<double> SampledDimension::axis(const ndsize_t count, const ndsize_t start
     double offset =  backend()->offset() ? *(backend()->offset()) : 0.0;
     double sampling_interval = backend()->samplingInterval();
     for (size_t i = 0; i < axis.size(); ++i) {
-        axis[i] = (static_cast<double>(i) + startIndex) * sampling_interval + offset;
+        double dbl;
+        if (!check::converts_to_double(i, dbl)) {
+            throw nix::OutOfBounds("SampledDimension::axis conversion to double failed!");
+        }
+        axis[i] = (dbl + startIndex) * sampling_interval + offset;
```
gicmo, line 262: Can you make the `check::converts_to_double` throw the exception? Like the other checks in there? More convenient I think.

jgrewe: as you wish :)

jgrewe:
This is good to go, I think. I will skip the column unit handling for now

PR #816, opened 2020-04-27
[Block/DataFrame] check for unique column names, ...

jgrewe:
added respective test

### DIFF
--- a/include/nix/Block.hpp
+++ b/include/nix/Block.hpp
@@ -23,6 +23,7 @@
 
 #include <string>
 #include <memory>
+#include <set>
 
 namespace nix {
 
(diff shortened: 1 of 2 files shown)


include/nix/Block.hpp
```diff
@@ -446,12 +446,21 @@ class NIXAPI Block : public base::EntityWithMetadata<base::IBlock> {
...
                 std::string msg = "Incompatible DataType for column ";
                 throw std::invalid_argument(msg + c.name);
             }
         }
+        std::sort(names.begin(), names.end());
+        names.erase(std::unique(names.begin(), names.end()), names.end());
+        if (names.size() != cols.size()) {
+            throw ConsistencyError("Block::createDataFrame: Column names must be unique!");
+        }
+
```
gicmo, line 463: I am ok with this but wouldn't using a [`set`](https://en.cppreference.com/w/cpp/container/set) be more convenient?

jgrewe: Yes, of course, thanks for pointing me to it!

gicmo:
Submission to codecov worked ([logs](https://travis-ci.org/github/G-Node/nix/jobs/680693447#L868) but codecov says its waiting for CI.

achilleas-k:
Force merging.

PR #825, opened 2020-07-26
[cmake] installation of headers also for static build

jgrewe:


achilleas-k:
Is this good?  Should I just merge?

PR #826, opened 2020-11-27
[dockerfile/arch] Fix aur package installation

achilleas-k:
Set PKGEXT to 'pkg.tar'
- Makes build slightly faster (no compression).

Use wildcards to install aur built packages:
- Makes dockerfile version independent, otherwise it would break every
  time one of the packages was updated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,9 @@ compiler:
 os:
   - linux
   - osx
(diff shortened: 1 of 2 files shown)


jgrewe (approved):

gicmo (approved):
:heart: awesome, thanks for that!

PR #828, opened 2020-11-27
[travis/macOS] disable brew update

jgrewe:


jgrewe:
travis is not starting with the latest change and it became messy, closing and coming back

